PR #2, opened 2020-01-20
Add specs.json with a single spec and linting

foolip:


README.md
```diff
@@ -3,3 +3,7 @@
 This repository contains a curated list of technical Web specifications that are directly implemented or that will be implemented by Web browsers.
 
 This list is meant to be an up-to-date input source for projects that run analyses on browser technologies to create reports on test coverage, cross-references, WebIDL, quality, etc.
+
+## Format
+
+`specs.json` contains an array of objects, each with a single property `url`.
```
tidoust: I understand that we'll have additional properties in some cases. That said, if a URL is going to be enough in most cases, wouldn't allowing simple strings on top of objects with a `url` property make the resulting file more readable (by humans that is, machines will first need to convert strings to objects with a `url` property to create a consistent list)?

foolip: I'd be fine with that, but then I'd probably want to treat the human readable list as input and have a build step to produce a uniform list so that tools need to do the work themselves. If you think that would be an improvement, can you file an issue?

PR #3, opened 2020-01-21
Allow specs to be represented by a simple string

tidoust:
In most cases, a URL should be all we need in that list. Getting rid of the wrapping `{ "url": ... }` struct in these cases allows to keep the list more human readable.

This update allows entries to be just a string. Linting code updated to prefer the string format when there is only a URL (which is going to be the case as long as we don't create additional properties). Specifications remain sorted by alphabetic order of their URL.

The `index.js` file continues to export an array of objects with a `url` property to ease ingestion by external tools.

See discussion in:
https://github.com/w3c/browser-specs/pull/2#pullrequestreview-345692753

(Note that if we actually discover later on that a URL is **not** enough in most cases, we may want to undo this change...)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,4 +9,5 @@ jobs:
       with:
         node-version: 12.x
     - run: npm ci
+    - run: npm run test
     - run: npm run lint
--- a/index.js
+++ b/index.js
@@ -1,5 +1,9 @@
 "use strict";
 
-const specs = require("./specs.json");
+const isString = obj =>
+  Object.prototype.toString.call(obj) === "[object String]";
+
+const specs = require("./specs.json")
+  .map(spec => isString(spec) ? { url: spec } : spec);
 
 module.exports = { specs };
--- a/index.js
+++ b/index.js
@@ -1,9 +1,6 @@
 "use strict";
 
-const isString = obj =>
-  Object.prototype.toString.call(obj) === "[object String]";
-
 const specs = require("./specs.json")
-  .map(spec => isString(spec) ? { url: spec } : spec);
+  .map(spec => (typeof spec === "string") ? { url: spec } : spec);
 
 module.exports = { specs };
--- a/lint.js
+++ b/lint.js
@@ -2,9 +2,6 @@
 
 const fs = require("fs").promises;
 
-const isString = obj =>
-  Object.prototype.toString.call(obj) === "[object String]";
-
 // Lint by normalizing specs.json and comparing it to the original,
 // fixing it in place if |fix| is true.
 async function lint(fix = false) {
@@ -12,7 +9,7 @@ async function lint(fix = false) {
   const specs = JSON.parse(specsBuffer);
 
   const sorted = specs
-    .map(spec => isString(spec) ? { url: spec } : spec)
+    .map(spec => (typeof spec === "string") ? { url: spec } : spec)
     .map(spec => Object.assign({}, spec, { url: new URL(spec.url).toString() }));
   sorted.sort((a, b) => a.url.localeCompare(b.url));
 
--- a/lint.js
+++ b/lint.js
@@ -1,6 +1,58 @@
 "use strict";
 
 const fs = require("fs").promises;
+const schema = require("./schema.json");
+const Ajv = require("ajv");
+const ajv = new Ajv();
+
+// When an entry is invalid, the schema validator returns one error for each
+// "oneOf" option and one error on overall "oneOf" problem. This is confusing
+// for humans. the following function improves the error being returned.
+const clarifyErrors = errors => {
+  if (!errors || (errors.length < 2)) {
+    return errors;
+  }
+
+  // If first two errors are type errors for oneOf choices, item is neither
+  // a string nor an object
+  if ((errors[0].schemaPath === "#/items/oneOf/0/type") &&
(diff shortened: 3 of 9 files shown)


index.js
```diff
@@ -1,5 +1,9 @@
 "use strict";
 
-const specs = require("./specs.json");
+const isString = obj =>
+  Object.prototype.toString.call(obj) === "[object String]";
+
+const specs = require("./specs.json")
+  .map(spec => isString(spec) ? { url: spec } : spec);
```
foolip: Why not just `typeof spec === "string"`? The input here is JSON so there can't be any String objects.

lint.js
```diff
@@ -2,19 +2,23 @@
 
 const fs = require("fs").promises;
 
+const isString = obj =>
```
foolip: `typeof spec === "string"` seems like it could replace this too.

lint.js
```diff
@@ -2,19 +2,23 @@
...
-    return { url };
-  });
-  fixed.sort((a, b) => {
-    a.url.localeCompare(b.url);
-  });
+  const specs = JSON.parse(specsBuffer);
+
+  const sorted = specs
+    .map(spec => isString(spec) ? { url: spec } : spec)
+    .map(spec => Object.assign({}, spec, { url: new URL(spec.url).toString() }));
```
foolip: This allows for arbitrary additional properties, most likely in the form of typos if we add more properties. I'd prefer this code to normalize everything that is known and drop anything else. Since currently no additional properties are allowed, just require all entries to be a string.

index.js
tidoust: Good point, fixed.

lint.js
tidoust: Done.

lint.js
tidoust: I was assuming that we would in any case add some kind of JSON schema to validate the structure before we pass it to the linter. I'll add some code logic to that effect.

foolip: I started out using ajv myself, but ripped it out since the schema was so trivial. Now when either a string or object are allowed it's less trivial, so I'm happy with using it.

I'll have a few days off for the lunar new year, but will review this when I'm back next week!

tidoust: I just added a JSON Schema and schema validation. The schema is a bit dumb for now since it allows an object with a URL, but then that's useless as long as we don't have additional properties, especially since the linting process will convert objects to URL strings. At least we'll have the right skeleton and logic in place for when we add properties.

tidoust:
Note I added schema validation and created tests for the linter.

.github/workflows/lint.yml
```diff
@@ -9,4 +9,5 @@ jobs:
       with:
         node-version: 12.x
     - run: npm ci
+    - run: npm run test
```
foolip: I'll send a PR to rename this workflow to reflect that it now does more than linting.

lint.js
```diff
@@ -1,27 +1,107 @@
...
+const ajv = new Ajv();
 
-// Lint by normalizing specs.json and comparing it to the original,
-// fixing it in place if |fix| is true.
-async function lint(fix = false) {
-  const specsBuffer = await fs.readFile("./specs.json");
-  let specs = JSON.parse(specsBuffer);
+// When an entry is invalid, the schema validator returns one error for each
+// "oneOf" option and one error on overall "oneOf" problem. This is confusing
+// for humans. the following function improves the error being returned.
```
foolip:
```suggestion
// for humans. The following function improves the error being returned.
```

lint.js
```diff
@@ -1,27 +1,107 @@
...
+const clarifyErrors = errors => {
+  if (!errors || (errors.length < 2)) {
```
foolip:
```suggestion
  if (!errors || errors.length < 2) {
```

lint.js
```diff
@@ -1,27 +1,107 @@
...
+  // a string nor an object
+  if ((errors[0].schemaPath === "#/items/oneOf/0/type") &&
+      (errors[1].schemaPath === "#/items/oneOf/1/type")) {
+    return [
+      Object.assign(errors[0], { "message": "should be a string or an object" })
+    ];
+  }
+
+  // Otherwise, if second error is a type error for second oneOf choice,
+  // it means the itme is actually a string that represents an invalid URL,
```
foolip:
```suggestion
  // it means the item is actually a string that represents an invalid URL,
```

lint.js
```diff
@@ -1,27 +1,107 @@
...
+  // which the first error should capture
```
foolip:
```suggestion
  // which the first error should capture.
```

specs.json
```diff
@@ -1,5 +1,3 @@
 [
-  {
-    "url": "https://compat.spec.whatwg.org/"
-  }
+  "http://compat.spec.whatwg.org/"
```
foolip: Should be `http://compat.spec.whatwg.org/`. I'll extend the lint to require this.

dontcallmedom: should be `https://compat.spec.whatwg.org/` rather, right?

.github/workflows/lint.yml
foolip: Actually, it still only concerns linting, I'll leave it for now.

specs.json
tidoust: Indeed, I missed the `s`. I guess it shows it's a good idea to add HTTPS enforcement to the linter...

PR #4, opened 2020-01-31
Fix nits/typos from "Allow specs to be represented by a simple string"

foolip:
https://github.com/w3c/browser-specs/pull/3


lint.js
```diff
@@ -7,9 +7,9 @@ const ajv = new Ajv();
 
 // When an entry is invalid, the schema validator returns one error for each
 // "oneOf" option and one error on overall "oneOf" problem. This is confusing
-// for humans. the following function improves the error being returned.
+// for humans. The following function improves the error being returned.
 const clarifyErrors = errors => {
-  if (!errors || (errors.length < 2)) {
+  if (!errors || errors.length < 2) {
```
tidoust: Do you have favorite code guidelines?

I don't mind either syntax, just wondering if this is more gut feeling or if you're following a specific rule here. For instance, if I run ESLint with `no-extra-parens`, it's going to report this line but it's also going to report other `if` lines below where parentheses could also be dropped.

foolip: My favorite style is any style that can be enforced and autofixed by tooling. These parens just seemed clearly unnecessary to me, but if a style rule added them back I wouldn’t mind.

PR #6, opened 2020-03-04
Detect and drop duplicate URLs when linting

tidoust:
It is relatively easy to end up with duplicate URLs if they get added in bulk.
The linter now only keeps the first one.


tidoust:
No-brainer, taking the liberty to merge.

PR #7, opened 2020-03-05
Add logic to compute shortname, family name and level

tidoust:
Most specification URLs follow a few common patterns from which it is easy to derive a shortname (e.g. "webauthn-2" from "https://www.w3.org/TR/webauthn-2/"). In turn, this shortname can be used to derive a level ("2" for "webauthn-2") and a family name that is common across levels of the spec ("webauthn" here).

The "compute-shortname.js" module does just that. It handles main types of URLs and returns an object with "shortname", "familyname", and a "level" property when that makes sense.

An exception is thrown when the shortname cannot be computed. When that happens, the corresponding entry in "specs.json" should be updated to make the shortname (and/or family name and/or level) explicit. For instance, the entry for CSS Media Queries Level 3 should typically read:

```json
{
  "url": "https://www.w3.org/TR/css3-mediaqueries/",
  "familyname": "mediaqueries",
  "level": 3
}
```

The list of specs exposed in index.js now features the "shortname", "familyname" and "level" properties for each spec.

The family name is typically the property that Reffy would use to name CSS/IDL extracts. Testing on the list of specs used by Reffy (not part of this update), this seems to work just fine.

The update comes with unit tests.

Additional notes:
- I'm not married to the term "familyname", it just seemed good to keep "shortname" for the leveled name, and I needed a name that carried some notion of group...
- I propose to move lint.js to the src folder in a follow-up PR, to only leave the list of specs and the index.js files (plus other boilerplate files) in the root folder.

### DIFF
--- a/src/compute-shortname.js
+++ b/src/compute-shortname.js
@@ -1,21 +1,22 @@
 /**
  * Module that exports a function that takes a URL as input and computes a
- * meaningful shortname, family name and level for it, when appropriate.
+ * meaningful name (i.e. the leveled shortname), shortname and level for it
+ * (when appropriate).
  *
- * The function returns an object with a "shortname" property. The shortname
- * matches the /TR/ shortname for specs published there. It includes the spec
- * level. For instance: "css-color-4" for "https://www.w3.org/TR/css-color-4/".
+ * The function returns an object with a "name" property. The name matches the
+ * /TR/ name for specs published there. It includes the spec level. For
+ * instance: "css-color-4" for "https://www.w3.org/TR/css-color-4/".
  *
- * For non-TR specs, the shortname returned is the "most logical" name that can
- * be extracted from the URL. The function typically handles a few typical cases
+ * For non-TR specs, the name returned is the "most logical" name that can be
+ * extracted from the URL. The function typically handles a few typical cases
  * (such as "https://xxx.github.io/" URLs). It throws an exception when no
- * meaningful shortname can be extracted.
+ * meaningful name can be extracted.
  *
- * Returned object will also alway have a "familyname" property that contains
- * an unleveled shortname for the specification. That name is shared across
- * levels of the specification. In most cases, it is the shortname without the
- * level. For instance: "css-page" for "https://www.w3.org/TR/css-page-4/".
- * In rare cases, note the familyname may be different. For instance:
+ * Returned object will also alway have a "shortname" property that contains
+ * an unleveled name for the specification. That shortname is shared across
+ * levels of the specification. In most cases, it is the name without its level
+ * suffix. For instance: "css-page" for "https://www.w3.org/TR/css-page-4/".
+ * In rare cases, note the shortname may be different. For instance:
  * "css-conditional" for "https://www.w3.org/TR/css3-conditional/".
  *
  * If the URL contains a level indication, the returned object will have a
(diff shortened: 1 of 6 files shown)


src/compute-shortname.js
```diff
@@ -0,0 +1,158 @@
+/**
+ * Module that exports a function that takes a URL as input and computes a
+ * meaninful shortname, family name and level for it, when appropriate.
```
dontcallmedom: s/meaninful/meaningful

Looking at [Version Management in W3C Technical Reports](https://www.w3.org/2005/05/tr-versions), the (arguably somewhat ambiguous) taxonomy seems to be that shortname is the non-leveled version; so maybe we should talk about "leveled shortname" and "shortname" (or "bare shortname", or...)

src/compute-shortname.js
```diff
@@ -0,0 +1,158 @@
...
+      return w3cTr[1];
+    }
+
+    // Handle WHATWG specs
+    const whatwg = url.match(/\/\/(.+)\.spec\.whatwg\.org\//);
+    if (whatwg) {
+        return whatwg[1];
+    }
+
+    // Handle extension specs defined in the same repo as the main spec
```
dontcallmedom: maybe quote an example?

src/compute-shortname.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    throw `Cannot extract meaningful shortname from ${url}`;
+  }
+
+  // Parse the URL to extract the shortname
+  const shortname = parseUrl(url);
+
+  // Make sure shortname looks legit
+  if (!shortname.match(/^[\w\-]+((?<=\-\d+)\.\d+)?$/)) {
```
dontcallmedom: add a human-readable interpretation of the regexp?

test/compute-shortname.js
```diff
@@ -0,0 +1,183 @@
...
+    it("handles extension specs defined in the same repo as the main spec (singular)", () => {
+      assertName("https://w3c.github.io/specwithext/extension.html", "specwithext-extension");
+    });
+
+    it("handles extension specs defined in the same repo as the main spec (plural)", () => {
+      assertName("https://w3c.github.io/specwithext/extensions.html", "specwithext-extensions");
+    });
+
+    it("handles CSS WG draft URLs", () => {
+      assertName("https://drafts.csswg.org/css-is-aweso/", "css-is-aweso");
```
dontcallmedom: ♥

src/compute-shortname.js
tidoust: Ah, I hadn't looked at that, good if we have some sort of official definition for `shortname`. I'll update the code accordingly. That would actually match the terminology used in Shepherd. And I can probably also follow Shepherd's naming convention for the "leveled shortname" and simply call the propery `name` (alternatives could be `leveledShortname` or perhaps `shortnameN` but neither looks appealing)

tidoust:
OK, code now uses and exposes the following properties:
- `name` for the leveled version of the name (e.g. "mediaqueries-4")
- `shortname` for the unleveled version of the name (e.g. "mediaqueries")
-  `level` for the level, which can also be a version number (e.g. 1.3)

tidoust:
Merging to be able to work on further features, notably support for the notion of "current" level and title retrieval. Logic may still be amended if needed.

Results are going to depend on the exact list of URLs that we start from. Testing with those used by Reffy and comparing with those defined in Shepherd, there will be a few discrepancies in generated names but, as far as I can tell, on top of a few missing specs (which is easy to fix), these discrepancies are due to Shepherd not always using names that match "reality" (in the sense of corresponding to an actual /TR/ or repo name). For instance, Shepherd has `permissions-1` but the actual name in /TR/ is `permissions`. To be discussed once the list gets created.

Issue #5, opened 2020-02-19
URL isn't sufficient information for useful processing


tabatkins:
Bikeshed uses its own bespoke set of specs to look at right now, preserved in a private db maintained by @plinss. It contains a little more information than just URLs, which Bikeshed uses in important ways, and so if Bikeshed and similar tools are meant to consume information from this list, presumably they'd need more info too.

Specifically, the relevant info that's tracked in Bikeshed's current db is:

* /TR url *and* draft/ED URL. (Bikeshed calls these "snapshot" and "current" urls, but I'm not picky about names.)
* shortname, both complete and (unversioned, version #) pair. (These two need to be separate because there's not a consistent pattern for formatting versioned shortnames; "svg2" versus "css-images-3", for example.)
* spec title; currently we track both the "full" title (like "CSS Images & Replaced Content Module Level 3") and a short title (like "CSS Images 3"); I use "full" in the bibliography index but "short" in section autolinks embedded in prose. For many specs these are identical.
* whether the spec is single-page or multi-page. The db tooling goes and discovers the relevant pages if it's a multipage spec; presumably we'd want to just put that info into the db.

plinss:
The spec DB also has the filename of the home page, which is usually Overview.html, but not always. Shepherd's spec parser needs that to understand that internal links to "spec/Overview.html" are actually links to "spec/" and prevents the home page from being parsed twice. I could hard-code 'Overview.html', 'index.html', and whatever else is in the DB at the moment, but I'd rather not.

In addition, the DB has the notion of the "current level" for a spec (which is often not just the last one), and an alternative shortname (which should actually be multiple alternatives), for when specs gets renamed so old URLs and other references can get redirected. Those fields are mostly used by the CSSWG draft server.

tidoust:
Pretty useful to have a vision of what info would be useful to have to make ingestion easy! Boundaries between projects are still fuzzy, I guess part of the whole exercise is about figuring out what these boundaries are and what are the sources of input.

I would separate information that can be automatically derived from information that needs to be manually entered. That is, I would view the `specs.json` file as an initial raw list of specs and would only put things in there that need to be manually maintained. From there, I would document how retrieve additional information about each spec. This repo could both contain tools that can retrieve that information, and the results of processing the initial list to have a list completed with basic information to ease ingestion by other projects (also see https://github.com/w3c/browser-specs/issues/1#issuecomment-576261400).

[...]

PR #8, opened 2020-03-15
Allow delta flag and add previous/next links

tidoust:
To make sense of the contents of a given spec, it is necessary to track whether it is a "delta" spec that may not contain the entire contents of the previous level or a "full" spec. This update allows specs to be defined as delta specs.

Whether the spec is a delta spec cannot be determined automatically. Delta specs need to be flagged with a "delta" flag in specs.json through a `delta` property as in:

```json
  {
    "url": "https://www.w3.org/TR/mydelta-2/",
    "delta": true
  }
```

To keep the list of URLs in specs.json readable, the flag can also be specified directly within the string that describes the spec, after the URL:

```
  "https://www.w3.org/TR/mydelta-2/ delta"
```

This notation is syntactic sugar for the previous notation and can only be used when a string is used to describe the specification. The linter will actually prefer the synctactic sugar notation when possible (meaning when no other property is defined).

To ease traversing operations between levels of the same specifications, the list of specs returned by "index.js" now features a `previousLevel` and `nextLevel` properties, when possible, set to the name of the spec that represents the previous and/or next level. By construction, all specs linked together in a chain of previous/next links share the same `shortname` property.

The `previousLevel` property is typically useful to identify the name of the previous level that contains the base contents of a delta spec.

From a code perspective:
- The linter was updated to handle the delta flag, notably the syntactic sugar notation, and to make sure that delta spec definitions are ok (delta specs should always have a previous non delta level in the list)
- A new compute-prevnext module was created to compute previous/next links
- The index.js code was updated to include previous/next links in the list of specs returned
- The JSON schema for specs.json was updated to add the "delta" flag and renamed to "specs-schema.json". The "multipleOf" condition was also dropped (because 0.1 multiples cannot work because JS only has floats).
- A new JSON schema was defined to check the list of specs returned by "index.js". It is similar to "specs-schema.json" but also has `previousLevel` and `nextLevel` properties.
- New tests were added for the linter, the new module, and for the list of specs returned by index.js.

### DIFF
--- a/index-schema.json
+++ b/index-schema.json
@@ -1,0 +1,39 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "type": "array",
+  "items": {
+    "type": "object",
+    "properties": {
+      "url": {
+        "type": "string",
+        "format": "uri"
+      },
+      "name": {
+        "type": "string",
+        "pattern": "^[\\w\\-]+((?<=\\-\\d+)\\.\\d+)?$"
+      },
+      "shortname": {
+        "type": "string",
+        "pattern": "^[\\w\\-]+$"
+      },
+      "level": {
+        "type": "number",
+        "minimum": 1
+      },
+      "delta": {
+        "type": "boolean"
+      },
+      "previousLevel": {
+        "type": "string",
+        "pattern": "^[\\w\\-]+((?<=\\-\\d+)\\.\\d+)?$"
+      },
+      "nextLevel": {
+        "type": "string",
+        "pattern": "^[\\w\\-]+((?<=\\-\\d+)\\.\\d+)?$"
+      }
+    },
+    "required": ["url", "name", "shortname"],
+    "additionalProperties": false
+  },
+  "minItems": 1
+}
--- a/index-schema.json
+++ b/index-schema.json
@@ -20,8 +20,9 @@
         "type": "number",
         "minimum": 1
       },
-      "delta": {
-        "type": "boolean"
+      "levelComposition": {
+        "type": "string",
+        "enum": ["full", "delta"]
       },
       "previousLevel": {
         "type": "string",
--- a/index.js
+++ b/index.js
@@ -1,9 +1,33 @@
 "use strict";
 
 const computeShortname = require("./src/compute-shortname.js");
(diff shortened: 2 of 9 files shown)


tidoust:
Not necessarily. It is useful to track "delta" specs in any case (we'll need that for IDL extracts), and having previous/next links to hop between levels of the same spec seems a good thing too. I'll work on the `currentLevel` property in a subsequent PR.

The one pending question is around the design of the `delta` flag as a Boolean. Thoughts?

dontcallmedom:
maybe instead of a boolean `delta`, `relationToPrevLevel` with possible values `replace` and `delta`?

tidoust:
So long for conciseness ;)

I was more looking for feedback on whether to change the boolean into an enum at all. What other value could we foresee on top of `replace` and `delta`? For instance, extensions are normally done using a different shortname, so `extend` does not seem like a good candidate.

dontcallmedom:
I guess the CSS mantra is prepare for the future, do not try to predict it, and hence an enum gives more room for extensibility than a boolean. I personally don't feel strongly about it, but the cost of adopting an enum sounds reasonably small.

plinss:
The TAG API design guidelines have similar opinions about bools in function calls.

tidoust:
I had checked the API design guidelines, but they only recommend to avoid unnamed bool arguments in functions, not to avoid bools in general. Anyway, let's switch to an enum.

Regarding the name of the property, I don't like `relationToPrevLevel`, although not for the same reason as @plinss in https://github.com/w3c/browser-specs/issues/5#issuecomment-599669900: the property will need to be set for all specifications and there are plenty of specs without previous levels.

The term `composition` seems too broad. It could easily be understood as describing what the spec contains (e.g. CSS, IDL, guidelines). That's info that could be interesting in itself, but that's orthogonal.

Would `levelComposition` be a good compromise?

Regarding possible values:
- `delta` seems good
- I don't like `replacement` for the same reason that some specs don't have previous levels and thus don't replace anything. What about `full` or `complete`?

tidoust:
OK, merging as-is. We can continue to bikeshed on names if needed, as long as we don't start exporting an actual list of specs

PR #21, opened 2020-03-28
 Add filtering param for index.js when used as CLI

tidoust:
The index of specifications is mostly aimed at being imported in other modules, but using it as a CLI is useful to visualize specs that the index contains. This update adds the ability to pass an identifier argument on the command-line that can either be:
- an integer, in which case it gets interpreted as the position of the spec of interest in the list. That is useful when a JSON schema validation error gets reported, since it often just points at a position in the list.
- a string, in which case the CLI returns all specs that have a URL, name, shortname or composition level that matches the string.

The CLI outputs a formatted spec object if there's only one spec that matches the filter argument, an array of spec objects if there's more than one.

PS: I need to clean my master branch locally, since it keeps on adding old commits to pull requests that have already been merged... Anyway, diff is correct, and not a big deal since I'll "squash and merge".

### DIFF
--- a/index-schema.json
+++ b/index-schema.json
@@ -24,6 +24,10 @@
         "type": "string",
         "enum": ["full", "delta"]
       },
+      "currentLevel": {
+        "type": "string",
+        "pattern": "^[\\w\\-]+((?<=\\-\\d+)\\.\\d+)?$"
+      },
       "previousLevel": {
         "type": "string",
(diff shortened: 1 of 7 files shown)


tidoust:
The update also drops the `forceCurrent` property that may appear in `specs.json` from the exported list: the info is exposed through the `currentLevel` property, and the JSON schema does not allow for the flag in the exported list in any case.

Issue #11, opened 2020-03-18
Expose TR/ED URLs


tidoust:
Most specs that are not "living standards" come into two variants: a spec in TR, and an Editor's Draft. Both are useful to track.

The TR URL should be the starting point for all specs that are published as TR documents.

Rules to determine the URL of an Editor's Draft:
1. Get it from the W3C API for TR specs
2. Get it from Specref for non-TR specs
3. Consider that the URL that serves as starting is the Editor's Draft (and that there is no TR spec in such cases)

Alternatively, we could perhaps start from the `name` of a spec and look it up in the W3C API, then in Specref. Problem is we would still have to use the URL for new specs that these systems won't know anything about.

Issue #10, opened 2020-03-18
Expose spec title


tidoust:
The spec title is obviously meant for humans. 3 ways to determine the title:

1. Get it from the W3C API for TR specs.
2. Get it from Specref for non-TR specs.
3. Fetch the spec and extract it from the HTML content

PR #22, opened 2020-03-31
Fetch spec info from W3C API, Specref or the spec

tidoust:
This update adds the logic needed to complete specification metadata with the spec's title (#10) and the TR/ED URLs (#11).

This additional information gets fetched from the W3C API when possible, from Specref when the W3C API does not know anything about the spec, or from the spec itself when neither the W3C API nor Specref knows anything about the spec.

Fetching logic is in `src/fetch-info.js`. By definition, fetching this information requires sending network requests and, as such, cannot be done in realtime on hundreds of specs. Instead, the commit adds a GitHub action set to run every 6 hours to fetch the information and update the `specs-info.json` file.

In turn, this file is imported by the `index.js` script, which now also returns the information.

New properties returned for each specification are:
- `title`: the title of the spec
- `trUrl`: the URL of the TR document for specs published in TR space
- `edUrl`: the URL of the latest Editor's Draft when known
- `source`: one of `w3c`, `specref`, or `spec` to give the source of the info

Other updates:
- Refactored JSON schemas in a schema folder with a common `dfns.json` file to avoid duplicating type definitions.
- Extended the filtering logic in `index.js` to also look for titles, source, and URLs.

Note that the code and the tests now expect to find a `config.json` file in the root folder that contains an object with a `w3cApiKey` property set to a valid W3C API key (this file is *not* part of the commit).

Also note that tests current fail if one tries to add a new spec to `specs.json` and does not run `npm run fetch-info`, because the new spec won't have a `title` property, which is defined as `required` in the JSON schema. This means that updates to `specs-info.json` need to be part of pull requests on the repo. Depending on the envisioned workflow for updates (and packaging), we might want to relax that rule later on.


tidoust:
Checks fail because linting workflow does not seem to have access to the `CONFIG_JSON` secret. This could be a feature if I read the doc correctly: "With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository":
https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#using-encrypted-secrets-in-a-workflow

Can we run tests on a pull request that need a secret? If not, we'll have to create a PR test suite that only runs tests that can be run without a W3C API key.

dontcallmedom:
A similar set up works on pull requests in [validate-repos](https://github.com/w3c/validate-repos/blob/master/.github/workflows/validate.yml), so I don't think the issue is with secrets not being available.

dontcallmedom:
when looking at the results of a run, it shows "***" where the secret should be in validate-repos, and nothing here.

tidoust:
> A similar set up works on pull requests in [validate-repos](https://github.com/w3c/validate-repos/blob/master/.github/workflows/validate.yml), so I don't think the issue is with secrets not being available.

But that set up is for a cron job and actual "push" actions, not for pull requests?

dontcallmedom:
no, it also works on pull requests, e.g. https://github.com/w3c/validate-repos/pull/86/checks?check_run_id=507710946

dontcallmedom:
hmm... but you're right that it's "not from a fork" (i.e. a local pull request only)

dontcallmedom:
maybe that's an acceptable limitation?

schema/dfns.json
```diff
@@ -0,0 +1,44 @@
+{
```
dontcallmedom, line 1: I initially thought (looking at the filename) that this was about definitions in specs (because `dfn` is the name of the HTML element), not shared definitions of for JSON schemata. Maybe rename it to definitions?

src/fetch-info.js
```diff
@@ -0,0 +1,264 @@
...
+    // be a spec with the same name published to /TR/ but that is probably an
+    // outdated version (e.g. WHATWG specs such as DOM or Fullscreen, or CSS
+    // drafts published a long long time ago)    
+    if (!spec.url.match(/^https?:\/\/(?:www\.)?w3\.org\/TR\/([^\/]+)\/$/)) {
+      return;
+    }
+
+    const url = `https://api.w3.org/specifications/${spec.name}`;
+    return new Promise((resolve, reject) => {
+      const request = https.get(url, options, res => {
```
dontcallmedom, line 61: any reason you're not using a higher level library (e.g. requests) for HTTP fetches?

.github/workflows/lint.yml
```diff
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: push
```
dontcallmedom, line 1: can we still run the w3c tests when the pull_request is local?

src/fetch-info.js
tidoust: Yes, I'm trying to introduce dependencies only when needed. We'll probably need to change that later on but network requests seem manageable with core node.js modules for now.

.github/workflows/lint.yml
tidoust: Well, if the pull request is local, it's a push, right?
That's actually the setting you have in `validate-repos`.

dontcallmedom: Ha, makes perfect sense indeed :)

Issue #14, opened 2020-03-18
Upload first list of specs


tidoust:
First list of specs should typically be:
- the result of merging `specs-css.json`, `specs-idl.json` and `specs-other.json` in [Reffy](https://github.com/tidoust/reffy/tree/master/src/specs)
- completed with specs in [Shepherd](https://api.csswg.org/shepherd/spec/?draft=1) that are not yet in these lists

tidoust:
Also see #27 for discussion on selection criteria.

PR #26, opened 2020-04-01
Add first list of specs

tidoust:
The list is based on the list used in Reffy, completed with a few additional specs from Shepherd.

I did not try to add all specifications for now. In particular, I did not add the following group notes that exist in Shepherd:
- [css-2007](https://drafts.csswg.org/css-2007/)
- [css-2010](https://drafts.csswg.org/css-2010/)
- [css-2015](https://drafts.csswg.org/css-2015/)
- [css-2017](https://drafts.csswg.org/css-2017/)
- [css-2018](https://drafts.csswg.org/css-2018/)
- [css-2020](https://drafts.csswg.org/css-2020/)
- [css-marquee-3](https://www.w3.org/TR/css3-marquee/)
- [css-mobile](https://drafts.csswg.org/css-mobile/)
- [css-preslev-1](https://drafts.csswg.org/css-preslev-1/)
- [css-print](https://drafts.csswg.org/css-print/)
- [css-template-1](https://drafts.csswg.org/css-template-1/)
- [fingerprinting-guidance-1](https://w3c.github.io/fingerprinting-guidance/)
- [matrix](https://drafts.fxtf.org/matrix/)

I also left out specs in Shepherd that look outdated:
- [css-page-template-1](https://drafts.csswg.org/css-page-template-1/)
- [css1](https://www.w3.org/TR/REC-CSS1/)
- [custom-1](https://drafts.fxtf.org/custom-1/)
- [dom-level-2-style](https://www.w3.org/TR/DOM-Level-2-Style/)
- [SVG 1.1](https://www.w3.org/TR/SVG11/)

... and did not add the following specs either:
- [json-ld11-api](https://www.w3.org/TR/json-ld11-api/) - Not implemented in browsers
- [css-egg-1](https://drafts.csswg.org/css-egg-1/) - April's Fool
- [css-viewport-1](https://drafts.csswg.org/css-viewport-1/) - Does not exist?

Linked to #14.


dontcallmedom:
I think the in/out question is one that likely will need solving sooner rather than later (I wonder if WPT has a clear criteria we could rely on?), but right now, I am unclear if this pull request is expected to the **the** list for the 1.0 release, or a first subset of the said list.

specs.json
```diff
@@ -1,3 +1,304 @@
...
+  "https://wicg.github.io/web-locks/",
+  "https://wicg.github.io/web-share-target/",
+  "https://wicg.github.io/web-transport/",
+  "https://wicg.github.io/webusb/",
+  {
+    "url": "https://www.khronos.org/registry/webgl/specs/latest/1.0/",
+    "name": "webgl1"
+  },
+  {
+    "url": "https://www.khronos.org/registry/webgl/specs/latest/2.0/",
```
dontcallmedom, line 86: is there any reason webgl2 is not described as a delta spec from webgl1?

(I'm not sure I understand if it even matters)

tidoust: I hadn't noticed that this was a delta spec in practice :)

It is different from other delta specs that we have, in that the "delta" status does not seem to be intented to disappear over time.

The "delta" flag impacts CSS/IDL extracts in Reffy: Reffy only produces these extracts for specs that are not flagged as "delta".

We may want to treat WebGL 1.0 and WebGL 2.0 as two separate specs in practice. If we flag the WebGL 2.0 as "delta", Reffy will only produce an IDL extract for WebGL 1.0. If we don't, Reffy will only produce an IDL for WebGL 2.0. Neither is what we need for WPT, which needs the two extracts (`webgl1.idl` and `webgl2.idl`, see https://github.com/web-platform-tests/wpt/tree/master/interfaces).

dontcallmedom: SGTM

tidoust:
I added several specs that already appear in Web Platform Tests, and created #27 to track selection criteria.

PR #40, opened 2020-04-11
Update URL of requestVideoFrameCallback spec

tidoust:
The spec used to be called `video.requestAnimationFrame()`. It is now called `HTMLVideoElement.requestVideoFrameCallback()` and its URL has changed.


tidoust:
No-brainer and data update job currently fails without it because old URL return a 404, so merging directly.

PR #42, opened 2020-04-30
Add Web Periodic Background Synchronization

tidoust:
Added to the crawl list in:
https://github.com/tidoust/reffy/commit/933defb8db699ea62631b1ecdee3556afc358baf


tidoust:
Note I created https://github.com/WICG/admin/pull/105 to update the URL of the spec on WICG side and thus in Specref

PR #48, opened 2020-05-25
Force Service Worker spec shortname

tidoust:
The list contains the nightly version of the spec, which makes sense but also generates a `ServiceWorker` shortname that should rather be `service-worker` (shortname used in /TR/).

This update forces the shortname to `service-worker`.


tidoust:
Hmm, I note en passant that I shouldn't have included index.json in the pull request...

PR #63, opened 2020-05-29
Update Background Sync URL

foolip:


foolip:
That is the goal :) Apologies to @stephenmcgruer who landed https://github.com/web-platform-tests/wpt/pull/23859 earlier today...

PR #65, opened 2020-06-01
Remove Web Budget API

foolip:
https://wicg.github.io/budget-api/ "is not being actively maintained"

The implementation was removed from Chromium:
https://groups.google.com/a/chromium.org/d/msg/blink-dev/18r3whCBv0I/b8qrtFTsDAAJ


foolip:
Thanks! I had also forgotten about the details of this spec, sorry for the churn :)

PR #68, opened 2020-06-03
Update Input Device Capabilities URL

foolip:
See https://github.com/WICG/admin/issues/112.


foolip:
Note that this URL doesn't work yet, this shouldn't be merged until the repo is renamed again.

Issue #62, opened 2020-05-28
Identify when a TR draft exists for a spec only known as editors draft


dontcallmedom:
When a spec goes to a FPWD, and assuming it was being tracked as an editor draft by browser-specs, we would want to update the entry in the list of specs to point to the TR draft.

Ideally, this would happen when running the crawl on editors draft - we should check if the TR url exists, and file an issue if it does.

Issue #71, opened 2020-06-11
Run script to find list of specs periodically


tidoust:
See discussion in https://github.com/w3c/browser-specs/pull/70#issuecomment-642508765

Goal is to set up a regular GitHub action to run the script that finds new specs periodically (once per week?), and have it raise a GitHub issue with new specs it may find.

Note that the action should also check `lastreviewed` dates in `monitor-repos.json` and force a periodic review for entries that have not been checked in the last 3 months or so.

dontcallmedom:
implemented in #76

PR #72, opened 2020-06-11
Add tests to compare base URL with release/nightly URL

tidoust:
The base URL should either match the release URL if there is one or the nightly URL otherwise. New tests will fail when there are entries in `index.json` for which that is not the case.

The first test should also help with identification of situations where a FPWD was published, and is known to Specref but not to us (#62).

Note that, as opposed to other tests that only depend on internal code logic, the second test may fail because external sources (Specref in practice) need to be fixed to return the right nightly URL. Fixing the problem at the source may take some time. Making tests fail certainly creates a strong incentive to fix the problem upstream, but means we're going to be stuck in a "some tests fail" situation in the meantime.

Typically, the tests currently choke on:
- `https://www.w3.org/TR/webxr-ar-module/`: the W3C API says that the release URL is `https://www.w3.org/TR/webxr-ar-module-1/` but then the right shortname for the W3C API remains `webxr-ar-module`. If we add `-1` to the base URL, the code won't be able to retrieve the info from the W3C API. Trying to see if this can be fixed in W3C systems.
- `https://www.w3.org/TR/webxr-gamepads-module/`: same reason
- `https://drafts.csswg.org/css-backgrounds-4/`: Specref returns `https://drafts.csswg.org/css4-background/` for the nightly URL. Filed https://github.com/w3c/csswg-drafts/pull/5194 to address this at the source
- `https://wicg.github.io/badging/`: URL needs to be updated to `https://w3c.github.io/badging/` in specs.json
- `https://wicg.github.io/geolocation-sensor/`: URL needs to be updated to `https://www.w3.org/TR/geolocation-sensor/` in specs.json
- `https://wicg.github.io/frame-timing/`: spec was published as a /TR/ Note but moved back to WICG. The URL of the ED URL needs to be updated in W3C systems (request sent).

PR should not be merged until these get fixed.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -86,4 +86,14 @@ describe("List of specs", () => {
     });
     assert.deepStrictEqual(wrong, []);
   });
+
+  it("uses the release URL as base URL when there is a release URL", () => {
+    const wrong = specs.filter(s => s.release && s.url !== s.release.url);
+    assert.deepStrictEqual(wrong, []);
+  });
+
+  it("uses the nightly URL as base URL when there is no release URL", () => {
+    const wrong = specs.filter(s => !s.release && s.url !== s.nightly.url);
+    assert.deepStrictEqual(wrong, []);
+  });
 });


tidoust:
I don't disagree. Let's leave this pull request pending while we figure out how to set things up for #71.

tidoust:
Superseded by #192

PR #76, opened 2020-06-12
Report specs found as candidates as github issue

dontcallmedom:
per #71

no idea if this will actually work, and not really sure how to test it beyond deploying it?

### DIFF
--- a/.github/candidate-specs.md
+++ b/.github/candidate-specs.md
@@ -1,6 +1,6 @@
 ---
 title: New specs for review
-assigneeds: tidoust, dontcallmedom
+assignees: tidoust, dontcallmedom
 labels: enhancement
 ---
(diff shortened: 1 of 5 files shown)


tidoust:
> no idea if this will actually work, and not really sure how to test it beyond deploying it?

One way to test it is to deploy it on a fork. But we can just as well deploy it here. At worst it just won't work.

dontcallmedom:
It actually [worked](#77)!

.github/candidate-specs.md
```diff
@@ -0,0 +1,10 @@
+---
+title: New specs for review
+assigneeds: tidoust, dontcallmedom
```
tidoust, line 3: I suspect s/assigneeds/assignees will fix the fact that the bot created an issue with assigning it to us.

dontcallmedom: that was a feature, but oh well

PR #82, opened 2020-06-16
Ignore specs not relevant to the scope of browser-specs 

dontcallmedom:
helps with #77


src/data/ignore.json
```diff
@@ -74,6 +74,27 @@
     },
     "w3c/webappsec-uisecurity": {
       "comment": "no longer worked on"
+    },
+    "w3c/dpub-aam": {
+      "comment": "not targeting browsers but epub readers"
+    },
+    "w3c/openscreenprotocol": {
+      "comment": "Protocol layer"
```
tidoust, line 82: It seems fine to ignore it for now but I'm torn on this one as it somewhat fits the "close to [the application layer]" criterion that we listed. That is, it is a protocol, but more an application protocol than a pure protocol.

dontcallmedom: one reason to include it may be that the definitions it has might be relevant to other specs (e.g. the related APIs); I'm happy to migrate it from ignore to specs.json if you feel that's more appropriate - I agree it's a bit of an edge case at the moment

tidoust: At some point, the mapping between the API and the protocol could move to the API, but right now it is done in the protocol spec. Perhaps that suggests that the repo should be put in the monitor list?

Issue #17, opened 2020-03-19
Add preview to pull requests


tidoust:
Once code is stable, most updates to the repository should be to add/modify the list of specs. To make it easy for people to contribute and for maintainers to review the PR, on top of running tests against the pull request, it would be good to automatically complete the PR with a preview of the information that the system would expose for the URLs modified.

tidoust:
Closing the issue as the new issue-based mechanism to suggest new specs includes such a preview. That does not work yet for pull requests, but that should also be less needed.

PR #90, opened 2020-06-16
Adding storage access and private click measurement as specs 

dontcallmedom:
Marking storage partitioning as ignorable

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -32,6 +32,8 @@
   "https://mathml-refresh.github.io/mathml-core/",
   "https://mimesniff.spec.whatwg.org/",
   "https://notifications.spec.whatwg.org/",
+  "https://privacycg.github.io/private-click-measurement/",
+  "https://privacycg.github.io/storage-access/",
   "https://quirks.spec.whatwg.org/",
   "https://storage.spec.whatwg.org/",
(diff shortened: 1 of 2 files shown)


dontcallmedom:
@tidoust is it to be expected that the test/lint fails on the pull request (but works on the push)?

tidoust:
> @tidoust is it to be expected that the test/lint fails on the pull request (but works on the push)?

No, I don't understand why the "push" version actually passes.

I do understand why the first one fails. This is not due to this pull request but to the previous commit: we now have 2 entries for `mediacapture-depth` because we actually had `https://www.w3.org/TR/mediacapture-depth/` already in the list, and the test correctly reports that as an error. Tests don't detect this on the pull request because they don't "build" the new index.json. They should, that is related to #17.

tidoust:
> No, I don't understand why the "push" version actually passes.

Hmm, I believe that the difference is that the `push` action runs with the repo code in the pull request (which includes a version of `index.json` without duplicates), whereas the `pull_request` action runs on the result of the merge between the code in the pull request and the `master` branch (and thus uses the latest version of `index.json` which has the duplicate code).

Actually, I don't think we should run the "Test and lint / lint (push)" action on pull requests, it should only run on `master`.

PR #92, opened 2020-06-16
Drop MediaStream Depth ED URL

tidoust:
Spec already covered through a /TR/ URL.
See discussion in https://github.com/w3c/browser-specs/pull/90#issuecomment-644755268


tidoust:
(Tests fail but that's normal since they run against the current `index.json` file)

dontcallmedom:
find-specs had flagged mediacapture-depth because the repo was using the wrong domain name for the editors draft (.com instead of .io)

PR #98, opened 2020-06-17
Mark a few more specs as to be monitored

dontcallmedom:
WebAppSec SubOrigins (dormant)
input for workers & purification (no clear implementation plans)


dontcallmedom:
s/wicg/WICG I notice now

PR #87, opened 2020-06-16
Add WICG specs with ongoing development in browsers

dontcallmedom:
Compared to #86, these implementations aren't quite public facing yet - I think they still match the criteria of implementors interest, but I thought their addition would benefit from separate explicit review.

https://wicg.github.io/content-index/spec/ from WICG/content-index
  origin trial in Chrome https://www.chromestatus.com/features/5658416729030656
https://wicg.github.io/custom-state-pseudo-class/ from WICG/custom-state-pseudo-class
  implemented in chrome behind a flag https://www.chromestatus.com/features/6537562418053120
https://wicg.github.io/overscroll-scrollend-events/ from WICG/overscroll-scrollend-events
  behind a flag per https://www.chromestatus.com/features/5650553247891456
https://wicg.github.io/savedata/ from WICG/savedata
  behind a flag https://www.chromestatus.com/features/5076813940260864
https://wicg.github.io/webpackage/loading.html
  origin trial in Chrome https://www.chromestatus.com/features/5126805474246656
https://wicg.github.io/is-input-pending/ from WICG/is-input-pending in development in Chrome per https://www.chromestatus.com/features/5719830432841728
https://wicg.github.io/local-font-access/ from WICG/local-font-access in development in Chrome per https://www.chromestatus.com/features/6234451761692672
https://wicg.github.io/webhid/ from WICG/webhid in development in Chrome per https://www.chromestatus.com/features/5172464636133376

PR #99, opened 2020-06-17
Add Media Feeds spec

dontcallmedom:
Has an implementation in Chrome behind a flag https://www.chromestatus.com/features/5695114963845120

this should probably be handled similarly to our approach to #86


dontcallmedom:
I meant to refer to #87 not #86 (since it is behind a flag) - but merging it as is also wfm :)

PR #100, opened 2020-06-17
Fix case-typo in monitored repo names

dontcallmedom:


tidoust:
I note that may be worth a test case, since that's the sort of things that can easily be detected but that we may not pay attention to...

dontcallmedom:
do you have a suggestion on how we would test this?

tidoust:
With a test (run on pull requests) that loads the ignore file and looks for entries that start with "wicg/"? We will likely put additional WICG specs in the ignore file over time, it seems worth some hardcoded test on it.

Issue #19, opened 2020-03-26
Expose spec repository


tidoust:
Tools may find it useful to get to the repository that hosts the specification. As with the spec title, 3 ways to determine the repo:

1. Get it from the W3C API for TR specs.
2. Get it from Specref for non-TR specs.
3. Fetch the spec and extract it from the HTML content, when possible

dontcallmedom:
The spec dashboard has existing logic to [map editors draft urls to github repo](https://github.com/w3c/spec-dashboard/blob/master/fetch-data/group-repos.js#L65) which might serve as inspiration here (not sure yet if the spec dashboard could en up reusing data or code from browser-specs)

foolip:
This would be useful for https://github.com/foolip/day-to-day/issues/85

foolip:
I've also written code to map URLs to repos, search for `specrepo ` in https://github.com/foolip/day-to-day/blob/a508cc63a9b2c1ec43603ede29cd24227038edf2/build/specs.js

tidoust:
Done and deployed in [v1.9.0](https://github.com/w3c/browser-specs/releases/tag/1.9.0) through a `repository` property under `nightly` (since the repository is associated with the nightly version)

PR #105, opened 2020-06-17
Expose spec repository in nightly.repository

tidoust:
See #19. The repository is either retrieved from Specref or computed from the nightly URL.

### DIFF
--- a/index.json
+++ b/index.json
@@ -8,7 +8,8 @@
       "currentSpecification": "compat"
     },
     "nightly": {
-      "url": "https://compat.spec.whatwg.org/"
+      "url": "https://compat.spec.whatwg.org/",
+      "repository": "https://github.com/whatwg/compat"
     },
     "title": "Compatibility Standard",
     "source": "specref"
@@ -22,7 +23,8 @@
       "currentSpecification": "console"
     },
     "nightly": {
-      "url": "https://console.spec.whatwg.org/"
+      "url": "https://console.spec.whatwg.org/",
+      "repository": "https://github.com/whatwg/console"
     },
     "title": "Console Standard",
     "source": "specref"
@@ -36,7 +38,8 @@
       "currentSpecification": "dom"
     },
     "nightly": {
-      "url": "https://dom.spec.whatwg.org/"
+      "url": "https://dom.spec.whatwg.org/",
+      "repository": "https://github.com/whatwg/dom"
     },
     "title": "DOM Standard",
     "source": "specref"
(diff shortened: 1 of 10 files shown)


index.json
```diff
@@ -679,7 +725,8 @@
       "currentSpecification": "trusted-types"
     },
     "nightly": {
-      "url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/"
+      "url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/",
+      "repository": "https://github.com/w3c/webappsec-trusted-types/"
```
dontcallmedom, line 729: Spurious trailing slash (none of the other repo urls have it)

index.json
```diff
@@ -749,7 +800,8 @@
       "currentSpecification": "background-fetch"
     },
     "nightly": {
-      "url": "https://wicg.github.io/background-fetch/"
+      "url": "https://wicg.github.io/background-fetch/",
+      "repository": "https://github.com/wicg/background-fetch"
```
dontcallmedom, line 804: not sure how to deal with it, nor how strongly we should care, but for WICG, WebAssembly, WebBluetoothCG, WebAudio, the mapping algorithm lowercase the path of the org name - github redirects the URL, but this will likely make reusing the data more difficult

index.json
tidoust: The inconsistency is in Specref. The code is enough to deal with all the nightly URLs that we have in the list for now, so we could perhaps just ignore repository info from Specref. Or we keep on using Specref and fix the info there, which would potentially create further hiccups in the future but would also improve alignment between projects.

dontcallmedom: if we don't need specref, I think we're better off limiting the dependency; fixing specref for that one wouldn't hurt though :)

index.json
```diff
@@ -1056,7 +1056,7 @@
     },
     "nightly": {
       "url": "https://wicg.github.io/frame-timing/",
-      "repository": "https://github.com/w3c/frame-timing"
+      "repository": "https://github.com/wicg/frame-timing"
```
dontcallmedom, line 1059: ha, this uncovered another bug - I guess in my check of repo existence, I should have checked they were not archived either

index.json
tidoust: Not sure how to deal with it either, I note specs are also inconsistent in the way they link to the repository. For instance, Background Fetch mentioned above uses `WICG` but [WebTransport](https://wicg.github.io/web-transport/) uses `wicg`.

dontcallmedom: I care more about the real name of the org than the way specs refer to it; we could either hardcode the names (there are only so many of them), or ask the info from the github API?

tidoust: Asking the info from the GitHub API would be nice. Is there a way to ask for all info at once? Otherwise, we'll end up with >300 requests to the API, which would exceed the [allowed limit](https://developer.github.com/v3/#rate-limiting) and require authentication.

dontcallmedom: I think we only need to ask about organization names, not repositories, so this should be about ~6 requests

tidoust: I rewrote the code to implement that logic. I note I didn't include tests on this specific part because, even with test that say send 3 requests to the GitHub API, it only takes 12 runs to be rate limited (and GitHub does enforce this limit).

Issue #12, opened 2020-03-18
Consider short titles


tidoust:
See discussion in https://github.com/w3c/browser-specs/issues/5#issue-567897765

It seems possible to compute a default shorter title (dropping words such as "Level" or "Module"), but that won't work in all cases. Ideally, that information would be defined in the spec, and would be tracked by the W3C API and Specref so that we don't have to do it here.

tabatkins:
Yeah, I'm happy to add a `Short Title` metadata to Bikeshed, which'll be communicated via a `<meta>` in the spec which can be picked up by the crawler.

In the absence of that, it looks like just literally dropping the words "Level", "Module", or "Specification" from the long title works out pretty great - I just tried it out and didn't spot a single spec title that is messed up by this.  (Possibly also "Standard", but that's only used by WHATWG specs and their titles tend to be short anyway, so we probably don't need to worry about it.)

So, with that heuristic applied by default, and an optional shortTitle key for the handful of specs that need something shorter (CSS has several where the full title is wordy), I think we could deploy this today, and then start relying on Bikeshed metadata afterward.

tabatkins:
Yeah, the Short Titles in Shepherd are manually maintained, it's just one of the fields you put in for the entry.

tidoust:
Comparing newly generated short titles in browser-specs with the list of short titles in Shepherd, I end up with the following differences (first title is the generated one in browser-specs, second the short title in Shepherd).

Need a manual entry in `specs.json`:
- [ ] [css-gcpm-4](https://drafts.csswg.org/css-gcpm-4/) - "CSS Generated Content for Paged Media 4" compared to "CSS GCPM 4"
- [ ] [css-multicol-2](https://drafts.csswg.org/css-multicol-2/) - "CSS Multi-column Layout 2" compared to "CSS Multicol 2"
- [ ] [css-size-adjust-1](https://drafts.csswg.org/css-size-adjust-1/) - "CSS Mobile Text Size Adjustment 1" compared to "CSS Size Adjustment 1"
- [ ] [compositing-2](https://drafts.fxtf.org/compositing-2/) - "Compositing and Blending 2" compared to "Compositing 2"
- [ ] [background-sync](https://wicg.github.io/background-sync/spec/) - "Web Background Synchronization" compared to "Background Sync"
[...]

PR #112, opened 2020-06-18
Add shortTitle property

tidoust:
See #12. The shortTitle property gets computed from the full title by dropping common terms and applying a few additional regular expressions. The shortTitle property may be specified in specs.json.

There should be ~25 specs for which we'll likely have to maintain the short title manually in specs.json. I haven't included them in the pull request so that we can first check that the automatic generation does not produce weird results.

### DIFF
--- a/index.json
+++ b/index.json
@@ -1000,7 +1000,7 @@
     },
     "title": "CORS and RFC1918",
     "source": "specref",
-    "shortTitle": "CORS and RFC1918 1918"
+    "shortTitle": "CORS and RFC1918"
   },
   {
     "url": "https://wicg.github.io/crash-reporting/",
--- a/index.json
+++ b/index.json
@@ -12,7 +12,8 @@
       "repository": "https://github.com/whatwg/compat"
     },
     "title": "Compatibility Standard",
-    "source": "specref"
+    "source": "specref",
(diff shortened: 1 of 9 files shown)


index.json
```diff
@@ -939,7 +999,8 @@
       "repository": "https://github.com/WICG/cors-rfc1918"
     },
     "title": "CORS and RFC1918",
-    "source": "specref"
+    "source": "specref",
+    "shortTitle": "CORS and RFC1918 1918"
```
dontcallmedom, line 1003: This is buggy

dontcallmedom:
FWIW, I used 
```js
const specs = require("./index.json");
console.log(JSON.stringify(specs.filter(s => s.title !== s.shortTitle).map(s => [s.title, s.shortTitle]), null, 2));
```
to compare `shortTitle` and `title`

tidoust:
Yes, I had done the same thing but failed to notice this one. I pushed an update, and simplified a tiny bit handling of non-breaking spaces (which `\s` actually matches).

FYI, css-backgrounds-4 is the spec with a non-breaking space in its title before the level. I don't know if we should preserve it, make it a default rule to always use a non-breaking space before the level number, or just replace with a space. Right now, I replace it with a regular space.

PR #114, opened 2020-06-18
Add script to assist in updating monitored specs

dontcallmedom:
The script checks the last-modified headers on the specs and highlights those that have changed since last review; it submits as a pull request (via a github action) the list of repos/specs to be monitored updated with a last review date of the day it is running on

See #113 for an example of what it produces.

### DIFF
--- a/.github/workflows/monitor-specs.yml
+++ b/.github/workflows/monitor-specs.yml
@@ -1,0 +1,28 @@
+name: Monitor specs
+
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


dontcallmedom:
that's a good point - I guess I was focusing on cases where the spec is the one that is lagging, but there are other circumstances worth watching for. Maybe we should classify what type of checks one is expected to do in the review (spec vs impl); right now, I see only one or two entries that aren't about the specs though.

PR #161, opened 2020-09-24
Add CSS Cascade 5

### LINKED ISSUE #158: New specs for review
[find-specs](src/find-specs.js) has identified the following candidates as potential new specs to consider:

- [ ] https://drafts.csswg.org/css-cascade-5/ from [w3c/csswg-drafts](https://github.com/w3c/csswg-drafts)
- [ ] https://wicg.github.io/responsive-image-client-hints/ from [WICG/responsive-image-client-hints](https://github.com/WICG/responsive-image-client-hints)

Please review if they match the inclusion criteria. Those that don't and never will should be added to [ignore.json](src/data/ignore.json), those that don't match yet but may in the future can be added to [monitor-repo.json](src/data/monitor-repos.json), and those that do match should be brought as a pull request on [specs.json](specs.json).

dontcallmedom:
close #158

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -6,6 +6,10 @@
   "https://drafts.css-houdini.org/font-metrics-api-1/",
   "https://drafts.csswg.org/css-animations-2/ delta",
   "https://drafts.csswg.org/css-backgrounds-4/ delta",
+  {
+    "url": "https://drafts.csswg.org/css-cascade-5/",
+    "shortTitle": "CSS Cascading 5"
+  },
   "https://drafts.csswg.org/css-env-1/",
   "https://drafts.csswg.org/css-extensions-1/",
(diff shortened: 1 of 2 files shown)


specs.json
```diff
@@ -179,7 +179,12 @@
   },
   {
     "url": "https://www.w3.org/TR/css-cascade-4/",
-    "shortTitle": "CSS Cascading 4"
+    "shortTitle": "CSS Cascading 4",
+    "forceCurrent": true
+  },
+  {
```
tidoust, line 185: Linter is picky and wants entries to be sorted by (otherwise opaque :)) URLs, so this entry needs to go near the top...

tidoust (approved):

PR #165, opened 2020-09-28
Handle new specs for review

### LINKED ISSUE #162: New specs for review
[find-specs](src/find-specs.js) has identified the following candidates as potential new specs to consider:

- [ ] https://w3c.github.io/merchant-validation/ from [w3c/merchant-validation](https://github.com/w3c/merchant-validation)
- [ ] https://wicg.github.io/file-system-access/ from [WICG/file-system-access](https://github.com/WICG/file-system-access) [chrome status](https://www.chromestatus.com/features/6284708426022912)

Please review if they match the inclusion criteria. Those that don't and never will should be added to [ignore.json](src/data/ignore.json), those that don't match yet but may in the future can be added to [monitor-repo.json](src/data/monitor-repos.json), and those that do match should be brought as a pull request on [specs.json](specs.json).

tidoust:
- Add Merchant validation spec to the monitor list (unofficial doc for now)
- Add File System Access to browser list (origin trial in Chrome)

Fixes #162

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -90,6 +90,7 @@
   "https://wicg.github.io/element-timing/",
   "https://wicg.github.io/entries-api/",
   "https://wicg.github.io/event-timing/",
+  "https://wicg.github.io/file-system-access/",
   "https://wicg.github.io/frame-timing/",
   "https://wicg.github.io/get-installed-related-apps/spec/",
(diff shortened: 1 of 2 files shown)


dontcallmedom (changes requested):
dup of #164 ; also filesystem is not a new api (but a renaming)

PR #166, opened 2020-09-29
Reflect transfer of WebTransport repo to w3c org

dontcallmedom:


tidoust (approved):
That's good but I note "someone" also needs to add the right redirect file to https://github.com/WICG/wicg.github.io to preserve the old URL of the Web Transport spec, coz' cool URIs don't change.

PR #169, opened 2020-09-30
Update URL of WebTransport spec

tidoust:
Repository dropped the `-` in its name.


tidoust:
FYI, I filed https://github.com/WICG/wicg.github.io/pull/10 to update the redirect accordingly.

dontcallmedom (approved):

PR #179, opened 2020-10-22
Drop Worklets spec

foolip:
https://drafts.css-houdini.org/worklets/ now redirects to HTML,
into which it's been integrated.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -493,6 +493,5 @@
     "url": "https://www.w3.org/TR/WOFF2/",
     "shortTitle": "WOFF 2.0"
   },
-  "https://www.w3.org/TR/worklets-1/",
   "https://xhr.spec.whatwg.org/"
 ]


foolip:
@tidoust I'll also need to manually remove the scraped data from webref, right?

tidoust:
I do see that the migration took place but don't see an internal ack for it, which I'm surprised about. Adding @dontcallmedom who has been following such migrations more closely.

dontcallmedom:
I think this should be merged (since the editors draft does redirect to HTML LS); I've followed up internally to make sure all the t get crossed.

dontcallmedom (approved):

tidoust:
> @tidoust I'll also need to manually remove the scraped data from webref, right?

Yes. I note that there are a few other specs to delete manually there following recent changes to browser-specs. I'll take care of that when I have a minute, feel free to beat me to it :)

PR #186, opened 2020-11-03
Get current specification from W3C API

### LINKED ISSUE #176: Get "current" level from W3C API
The code that determines the current specification in a series currently considers that the current level is the last level that is not a delta spec, unless otherwise specified in `specs.json`:
https://github.com/w3c/browser-specs/blob/master/src/compute-currentlevel.js#L11

For /TR/ specs, the "current" spec should be the level that gets returned when `/TR/[series shortname]` is fetched. We can probably use that (or the info in [Shepherd](https://drafts.csswg.org/)) as the source of truth. That would have the advantage of making it easier to follow transitions that are doomed to happen once in a while.

See related discussion in https://github.com/tidoust/reffy/issues/417#issuecomment-707701186

tidoust:
Fixes #176.

The current specification in a series of specifications was the last level that was not a delta spec, unless otherwise specified in specs.json.

This update extends the definition as follows: for W3C specifications, the current specification becomes the one returned by the W3C API, unless otherwise specified in specs.json, and provided that the current specification returned by the W3C API is in the list.

In turn, this allows us to drop the manual "current" entries in specs.json. They are no longer needed, the W3C API returns the right info for them.

This will trigger one change in the resulting index: the current specification for Compositing and Blending will become Level 1. It used to be Level 2, but Level 2 has not been published as a /TR/ draft yet, so it makes sense to me to have Level 1 as the current specification. I note that this is not fully aligned with what the unversioned ED URL returns though, since https://drafts.fxtf.org/compositing/ returns Level 2.

One possible update that we could consider is to make sure that, for /TR/ specs, we always have the current specification in browser-specs. This would require adding the following specifications that are the current ones of more recent ones in browser-specs:
- [`accname-1.1`](https://www.w3.org/TR/accname-1.1/)
- [`core-aam-1.1`](https://www.w3.org/TR/core-aam-1.1/)
- [`hr-time-2`](https://www.w3.org/TR/hr-time-2/)
- [`navigation-timing`](https://www.w3.org/TR/navigation-timing/)
- [`pointerevents2`](https://www.w3.org/TR/pointerevents2/)
- [`pointerlock`](https://www.w3.org/TR/pointerlock/)
- [`resource-timing-1`](https://www.w3.org/TR/resource-timing-1/)
- [`wai-aria-1.1`](https://www.w3.org/TR/wai-aria-1.1/)
- [`webauthn-1`](https://www.w3.org/TR/webauthn-1/)
- [`WOFF`](https://www.w3.org/TR/WOFF/)

I'm not convinced that's really needed.

### DIFF
--- a/src/build-index.js
+++ b/src/build-index.js
@@ -88,22 +88,22 @@ const specs = require("../specs.json")
 
 
 // Fetch additional spec info from external sources and complete the list
-// Note on the "assign" call:
-// - `{}` is needed to avoid overriding spec
-// - `spec` appears first to impose the order of properties computed above in
-// the resulting object
-// - `specInfo[spec.shortname]` is the info we retrieved from the source
-// - final `spec` ensures that properties defined in specs.json override info
-// from the source.
 fetchInfo(specs, { w3cApiKey })
   .then(specInfo => specs.map(spec => {
+    // Note on the "assign" call:
+    // - `{}` is needed to avoid overriding spec
+    // - `spec` appears first to impose the order of properties computed above
+    // in the resulting object
+    // - `specInfo[spec.shortname]` is the info we retrieved from the source
+    // - final `spec` ensures that properties defined in specs.json override
+    // info from the source.
     const res = Object.assign({}, spec, specInfo[spec.shortname], spec);
 
     // Update the current specification based on the info returned by the
     // W3C API, unless specs.json imposed a specific level.
     // Note: the current specification returned by the W3C API may not be in the
     // list, since we tend not to include previous levels for IDL specs (even
-    // if they are still "current"), in which case, we'll just ignore the info
+    // if they are still "current"), in which case we'll just ignore the info
     // returned from the W3C API.
     const currentSpecification = specInfo.__current ?
       specInfo.__current[spec.series.shortname] : null;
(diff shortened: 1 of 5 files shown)


src/fetch-info.js
```diff
@@ -115,13 +114,59 @@ async function fetchInfoFromW3CApi(specs, options) {
...
         title: info[idx].title
       };
     }
   });
+
+  // Fetch info on the series
+  const seriesShortnames = [...new Set(
+    Object.values(results)
+      .map(spec => spec.series)
+      .map(series => series.shortname))];
```
dontcallmedom, line 134: why not `spec => spec.series.shortname` directly?

src/build-index.js
```diff
@@ -96,8 +96,26 @@ const specs = require("../specs.json")
 // - final `spec` ensures that properties defined in specs.json override info
 // from the source.
 fetchInfo(specs, { w3cApiKey })
-  .then(specInfo => specs.map(spec =>
-    Object.assign({}, spec, specInfo[spec.shortname], spec)))
+  .then(specInfo => specs.map(spec => {
+    const res = Object.assign({}, spec, specInfo[spec.shortname], spec);
```
dontcallmedom, line 100: now that we don't just return `Object.assign`, the comment describing it above should be moved in-line.

dontcallmedom (approved):
LGTM with a couple of nits

PR #191, opened 2020-11-09
Fix relative link to find-specs script

tidoust:
Link is currently broken, see for example https://github.com/w3c/browser-specs/issues/188


tidoust:
Wait, there are more links to fix :)

PR #192, opened 2020-11-09
Add periodic check on base URL

tidoust:
The base URL of a specification should either match its release URL if there is one or its nightly URL otherwise.

This update adds a `check-base-url` script and a weekly GitHub action, run on Mondays at 00:30 (to avoid collision with the monitor action), to make sure that this is actually the case. The action creates an issue when there are specifications to check.

Cancels and replaces #72.

### DIFF
--- a/src/check-base-url.js
+++ b/src/check-base-url.js
@@ -1,0 +1,28 @@
+/**
+ * CLI tool that parses the generated index of specifications to make sure that
+ * the base URL either matches the release URL if there is one, or the nightly
+ * URL otherwise.
+ *
+ * The CLI tool returns Markdown that can typically be used to create an issue.
+ * It also sets a check_list environment variable that can be used in GitHub
+ * actions.
+ *
+ * No content is returned when everything looks good.
+ */
+
+const core = require("@actions/core");
+const specs = require("../index.json");
+
+const problems = specs
+  .filter(s => (s.release && s.url !== s.release.url) || (!s.release && s.url !== s.nightly.url))
+  .map(s => {
+    const expected = s.release ? "release" : "nightly";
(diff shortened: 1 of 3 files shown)


tidoust:
Note the script currently (correctly) reports that the CSS21 entry is problematic because the release URL is [https://www.w3.org/TR/CSS2/](https://www.w3.org/TR/CSS2/) whereas the base URL is [https://www.w3.org/TR/CSS21/](https://www.w3.org/TR/CSS21/). Request sent to have the release URL updated in W3C database.

src/check-base-url.js
```diff
@@ -0,0 +1,28 @@
...
+
+const core = require("@actions/core");
+const specs = require("../index.json");
+
+const problems = specs
+  .filter(s => (s.release && s.url !== s.release.url) || (!s.release && s.url !== s.nightly.url))
+  .map(s => {
+    const expected = s.release ? "release" : "nightly";
+    const expectedUrl = s.release ? s.release.url : s.nightly.url;
+    return `- [ ] [${s.title}](${s.url}): expected ${expected} URL [${expectedUrl}](${expectedUrl}) but base URL is [${s.url}](${s.url})`;
```
dontcallmedom, line 21:
```suggestion
    return `- [ ] [${s.title}](${s.url}): expected ${expected} URL ${expectedUrl} but base URL is ${s.url}`;
```
URLs get auto-linked IIRC

tidoust:
> Note the script currently (correctly) reports that the CSS21 entry is problematic because the release URL is https://www.w3.org/TR/CSS2/ whereas the base URL is https://www.w3.org/TR/CSS21/. Request sent to have the release URL updated in W3C database.

W3C API updated, the script should no longer report any failure.

dontcallmedom (approved):

PR #195, opened 2020-11-26
Remove Trace Context

dontcallmedom:
Does not match the browser-spec requirements

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,8 @@ jobs:
         npm ci
     - name: Build new index file
       run: npm run build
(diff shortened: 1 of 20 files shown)


dontcallmedom:
@foolip it looks you added the spec back in the days the spec list was maintained in reffy: https://github.com/foolip/reffy/commit/e80974c5aeb9e9991632ca2096973756e2518f65 - was there a specific motivation for having it there?

tidoust (approved):
I propose to take silence as consent. The spec is not in Web Platform Tests and there are no known references to it from other specs for now.

PR #196, opened 2020-11-26
Add ECMA 262 and TC39 Stage 3 proposals to the list of specs 

dontcallmedom:
do not merge for now - reffy is not ready to work with github markdown pages as "specs".

### DIFF
--- a/src/compute-shortname.js
+++ b/src/compute-shortname.js
@@ -56,11 +56,6 @@ function computeShortname(url) {
         return whatwg[1];
     }
 
-    // Handle ECMAScript
-    if (url === "https://tc39.es/ecma262/") {
-      return "ecmascript";
-    }
-
     // Handle TC39 Proposals
     const tc39 = url.match(/\/\/tc39\.es\/proposal-([^\/]+)\/$/);
     if (tc39) {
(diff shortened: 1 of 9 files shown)


dontcallmedom:
now ready

src/compute-shortname.js
```diff
@@ -56,6 +56,18 @@ function computeShortname(url) {
         return whatwg[1];
     }
 
+    // Handle ECMAScript
+    if (url === "https://tc39.es/ecma262/") {
+      return "ecmascript";
+    }
```
tidoust, line 62: Given that this only deals with one spec, could that appear in a `shortname` property in `specs.json` instead?

tidoust (commented):
See inline.

PR mentions that Reffy is not ready to work with GitHub markdown pages as specs. Was it solved by restricting the list to "stage 3" proposals that don't use markdown?

tidoust:
I added a couple of tests. I don't see any direct way to test retrieval of the title: I suppose TC39 proposals are not meant to stay as proposals forever so they don't seem to be good candidate URLs for tests, and the title for the main ECMAScript spec actually comes from Specref.

dontcallmedom:
yes, with stage 3 only proposals, reffy seems to work reasonably well.

tidoust (approved):

PR #204, opened 2020-12-07
Reflect specs identified in December 7 spec finder

### LINKED ISSUE #202: New specs for review

- [ ] https://tc39.es/proposal-hashbang/ from [tc39/proposal-hashbang](https://github.com/tc39/proposal-hashbang)
- [ ] https://tc39.es/proposal-regexp-legacy-features/ from [tc39/proposal-regexp-legacy-features](https://github.com/tc39/proposal-regexp-legacy-features)
- [ ] https://wicg.github.io/document-policy/ from [WICG/document-policy](https://github.com/WICG/document-policy) [chrome status](https://www.chromestatus.com/features/5645593894453248)
- [ ] https://www.w3.org/TR/webxrlayers-1/ from [immersive-web/layers](https://github.com/immersive-web/layers)

dontcallmedom:
close #202

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -44,7 +44,6 @@
   "https://immersive-web.github.io/anchors/",
   "https://immersive-web.github.io/dom-overlays/",
   "https://immersive-web.github.io/hit-test/",
-  "https://immersive-web.github.io/layers/",
   "https://infra.spec.whatwg.org/",
   "https://mathml-refresh.github.io/mathml-core/",
   "https://mimesniff.spec.whatwg.org/",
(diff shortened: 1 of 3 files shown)


tidoust (approved):

tidoust (approved):
I note that I am not entirely clear why some repos/specs need to be under `repos`, while others need to be under `specs`. Is there a logical rule to select one or the other when both are in theory possible?

dontcallmedom:
the distinction is indeed far from obvious, nor ideal - in some cases, the spec finder starts from a list of repos, in which case it filters them through the repo angle first, then through spec urls; in others, it starts from a list of links, in which case it only looks at spec urls.

PR #207, opened 2020-12-08
Adopt new TR shortname for CSS Conditional 3

tidoust:
`css3-conditional` now redirects to `css-conditional-3`

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -246,6 +246,10 @@
   "https://www.w3.org/TR/css-color-5/ delta",
   "https://www.w3.org/TR/css-color-adjust-1/",
   {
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
how did you notice it? (wondering if that's something we can/should detect automatically)

tidoust:
The build started to [break](https://github.com/w3c/browser-specs/runs/1515917661) today :)

The code detects that the W3C API returns a redirect, complains that the shortname should be updated, and fails. In theory, it would be possible to improve the code to rather follow the redirect, raise an issue (or even a PR), and pass, but at least we're able to detect the situation automatically... and we get some motivation to act on it.

PR #212, opened 2020-12-14
Fix JSON in ignore list

dontcallmedom:


tidoust (approved):
It should be prett easy to detect such situations automatically, I created #214 to track that.

PR #216, opened 2020-12-14
Avoid running tests twice on internal pull requests

tidoust:
There were two actions that ran tests: one that reacted to `push` and that ran all tests, and another that reacted to `pull_request` and that only ran tests that do not need a W3C API token, since that secret is inaccessible from PRs created from another repository.

One net result is that tests ran twice for pull requests created from this repository (because they are both a `push` and a `pull_request`).

This update gets back to only one action and triggers the right set of tests depending on whether the action can access the secret. This ensures that we run as many tests as possible against pull requests, and only run them once.


tidoust:
Wait, that wouldn't de-deduplicate jobs, actually, need to be smarter than that.

tidoust:
That should now be correct if I understand things correctly. The tests are going to run when we push against `master` and when we create a PR against the `master` branch. They won't run if we push to a different branch, but we don't really need that.

.github/workflows/lint.yml
```diff
@@ -1,14 +1,33 @@
...
+          with:
+            node-version: 12.x
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Test (including W3C API tests)
+        if: ${{ secrets.CONFIG_JSON }}
+        run: |
+          echo "${{ secrets.CONFIG_JSON }}" | base64 --decode > config.json
```
dontcallmedom, line 25: Instead of putting JSON-encoded strings in secrets, I liked the way @foolip [did it in validate repo](https://github.com/w3c/validate-repos/blob/master/.github/workflows/validate.yml#L23)

tidoust: Right but this is not the PR you're looking for :)

That is not what that I am trying to fix here. Plus that part works...

dontcallmedom: fair enough :)

dontcallmedom (approved):

PR #217, opened 2020-12-14
Fix test GitHub action

tidoust:
There is apparently no way to access secrets from within an `if:` statement.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,15 +22,12 @@ jobs:
         run: npm ci
 
       - name: Test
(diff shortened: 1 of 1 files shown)


tidoust:
Grrr, does not work for now :(

tidoust:
Yeah, nailed it. So no way to access secrets from an `if:` but one can set it as an env variable and reference it from an `if:`.
@dontcallmedom, FYI. I'm taking the liberty to merge to fix the action. I also took the liberty to fix the build action that was also broken :(

PR #218, opened 2020-12-15
Add tests to validate JSON files in data folder

### LINKED ISSUE #214: Add test to validate JSON files in data folder
See https://github.com/w3c/browser-specs/pull/212

tidoust:
Tests also make sure that there cannot be an entry that is both in the ignore list and in the monitor list.

Note the check on repository names may be a bit too constrained as one can probably use other characters in repo names, but then all repos pass this for now.

Close #214.

### DIFF
--- a/test/data.js
+++ b/test/data.js
@@ -1,0 +1,65 @@
+/**
+ * Make sure that the src/data/*.json files respect the right JSON schema
+ */
+
+const assert = require("assert");
+const schema = require("../schema/data.json");
+const dfnsSchema = require("../schema/definitions.json");
+const Ajv = require("ajv");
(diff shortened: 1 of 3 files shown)


test/data.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * Make sure that the src/data/*.json files respect the right JSON schema
+ */
+
+const assert = require("assert");
+const schema = require("../schema/data.json");
```
dontcallmedom, line 6: schema/data.json should be part of this PR :)

tidoust: It should :)

dontcallmedom (approved):

PR #219, opened 2020-12-17
Mark CSS Images 4 as a delta spec

dontcallmedom:
Per https://github.com/w3c/csswg-drafts/issues/5804#issuecomment-747420357
Helps with https://github.com/w3c/webref/issues/77

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -295,6 +295,7 @@
   },
   {
     "url": "https://www.w3.org/TR/css-images-4/",
+    "seriesComposition": "delta",
     "shortTitle": "CSS Images 4"
   },
   "https://www.w3.org/TR/css-inline-3/",


LeaVerou (approved):

tidoust:
I believe that this will not be enough to solve the initial problem in #77, because `currentSpecification` would remain Level 4 given that this is what the W3C API currently asserts.

We may either wait for that to be fixed in the W3C API (see https://github.com/w3c/csswg-drafts/issues/5804#issuecomment-747449515) or add a `"forceCurrent": true` property in `specs.json` for the Level 3 entry.

dontcallmedom:
@tidoust now that /TR has been updated, can we merge?

PR #220, opened 2020-12-17
Add ECMAscript Internationalization API spec

tidoust:
The `ecma-402` shortname is based on that used in Specref.
Short title may be too short, perhaps?

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -70,6 +70,11 @@
     "shortname": "ecmascript",
     "shortTitle": "ECMAScript"
   },
+  {
+    "url": "https://tc39.es/ecma402/",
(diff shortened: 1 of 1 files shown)


dontcallmedom:
yeah, I think the shortTitle is too short; Ecmascript Internationalization sounds like short enough to me.

tidoust:
> yeah, I think the shortTitle is too short; Ecmascript Internationalization sounds like short enough to me.

Then not need to specify it manually. The short title that gets computed automatically is "ECMAScript Internationalization API", which seems good to me.

dontcallmedom (approved):

PR #227, opened 2020-12-23
Expose test repository and paths for each spec

### LINKED ISSUE #201: Manage and expose related WPT folder
It would be good to link specs to their test suite in WPT.

Most of that mapping info can likely be determined automatically by looking at `META.yml` files in WPT, and/or by searching for a link to the test suite in the spec itself.

Practically speaking, a script that runs weekly and creates PR automatically should do the trick.

tidoust:
This update exposes a new `tests` property that details, when that is possible:
1. the URL of the repository that contains the test suite for the spec
2. the actual paths in that repository that contain the test suite
3. if needed, the sub-paths to exclude from the list of paths because they actually contain tests for another specification

@foolip That information should allow you to drop all `testrepo` and `testpath` entries in [day-to-day/specs-fixes.json](https://github.com/foolip/day-to-day/blob/main/specs-fixes.json) and associated logic. FWIW, the list of folders for HTML in that file does not seem fully correct anymore (`shadow-dom` should rather be associated with DOM, `2dcontext`, `innerText` and `offscreen-canvas` no longer exist, and `x-frame-options` should also be in the list).

A few exceptions need to be handled in specs.json:
- The ECMAScript spec
- WebGL 1 and 2
- The WebAssembly Core spec, which uses its own repository
- The WebGPU spec. The spec does have a folder under WPT but it is empty for now as the actual test suite is developed in the gpuweb/cts repository.

A few notes:
- The information is based on `META.yml` files in the Web Platform Tests (WPT) repository.
- The code associates WebGL extension specs with the main WebGL test repository. The association could be more fine-grained.
- Similarly, the code skips TC39 proposals for now, which may or may not have tests in the main ECMAScript test repository.
- The code clones the WPT repository to a local cache folder because the repo is too large to be processed using the GitHub API (we'd be quickly rate limited, even with an API token). To avoid creating too many files, the checkout is restricted to META.yml files.
- `testPaths` is an array that contains one entry in most cases, except for DOM, HTML and a couple of other specs. An alternative would be to use a space- or comma-separated list as in foolip/day-to-day but then that would require parsing the values.
- The whole thing cannot deal with cases where a test suite is composed of tests defined in multiple repositories. That should not happen, although I note that there are apparently ECMAScript tests in the [`js`](https://github.com/web-platform-tests/wpt/tree/master/js) WPT folder which I'm ignoring for the time being...
- The code does not include provisions to detect changes. For instance, it would seem useful to review the ~50 specs that don't have a test suite once in a while to make sure that they still don't have one. To be done separately.

This would close #201.

### DIFF
--- a/src/determine-testpath.js
+++ b/src/determine-testpath.js
@@ -137,18 +137,7 @@ module.exports = async function (specs, options) {
     };
 
     if (spec.tests) {
-      if (spec.tests.repository) {
-        info.repository = spec.tests.repository;
-      }
-
-      if (spec.tests.testPaths) {
-        const testPaths = spec.tests.testPaths;
-        info.testPaths = testPaths;
-        if (spec.tests.excludePaths) {
-          info.excludePaths = spec.tests.excludePaths;
-        }
-      }
-      return info;
+      return Object.assign(info, spec.tests);
     }
 
     if (spec.url.startsWith("https://www.khronos.org/")) {
--- a/src/determine-testpath.js
+++ b/src/determine-testpath.js
@@ -1,0 +1,210 @@
+/**
+ * Module that takes a list of spec objects as input and returns, for each spec,
+ * the URL of the repository that contains the test suite of the spec along with
+ * the path under which the tests are to be found in that repository.
+ *
+ * The function will run git commands on the command-line and populate the local
+ * ".cache" folder.
+ */
+
+const fs = require("fs");
+const path = require("path");
+const execSync = require("child_process").execSync;
+
+// Cache folder under which the WPT repository will be cloned
+const cacheFolder = path.resolve(__dirname, "..", ".cache");
+const wptFolder = path.resolve(cacheFolder, "wpt");
+
+/**
+ * Helper function to setup the cache folder
+ */
+function setupCacheFolder() {
+  try {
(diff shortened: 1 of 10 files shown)


src/determine-testpath.js
```diff
@@ -0,0 +1,210 @@
...
+
+  const wptFolders = await readWptMetaFiles();
+
+  function determineTestInfo(spec) {
+    const info = {
+      repository: "https://github.com/web-platform-tests/wpt"
+    };
+
+    if (spec.tests) {
+      if (spec.tests.repository) {
```
dontcallmedom, line 140: couldn't this whole block be simplified in 
`return Object.assign({}, spec.tests);`?

dontcallmedom (approved):
brilliant :)

src/determine-testpath.js
tidoust: More `return Object.assign(info, spec.tests);` as we'll want to use the WPT repo by default unless otherwise specified in specs.json, but indeed, that can be simplified.

tidoust: Well, `repository` is mandatory for now under `tests` in `specs.json`, so your code would work as well in practice, but we might want to relax that rule in the future, so I'll continue to extend `info`.

PR #232, opened 2021-01-18
Always check repo against ignore/monitor lists

tidoust:
The tool that looks for potential new specs did not always check candidate repos against the list of repos in `ignore.json` and `monitor.json`. This update forces the code to look at these lists when it considers whether a candidate is in scope.

Note explicit checks on the initial list of repos are no longer stricto senso necessary and have been dropped.

Done to address #229, which should not have contained EPUB entries because the underlying repo is in the ignore list.

### DIFF
--- a/schema/data.json
+++ b/schema/data.json
@@ -4,6 +4,22 @@
 
   "type": "object",
   "properties": {
+    "groups": {
+      "type": "object",
+      "propertyNames": {
+        "type": "string"
+      },
+      "additionalProperties": {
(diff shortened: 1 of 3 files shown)


dontcallmedom (approved):
LGTM - maybe also add "EPUB 3 Working Group" to the list of WGs to skip?

tidoust:
Done and I also moved the list to `ignore.json`.

dontcallmedom (approved):
even better :)

PR #233, opened 2021-01-23
Add DataCue API

foolip:
This is implemented in Safari, and was also in Edge 18.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -136,6 +136,7 @@
   "https://wicg.github.io/crash-reporting/",
   "https://wicg.github.io/css-parser-api/",
   "https://wicg.github.io/custom-state-pseudo-class/",
+  "https://wicg.github.io/datacue/",
   "https://wicg.github.io/deprecation-reporting/",
   "https://wicg.github.io/element-timing/",
   "https://wicg.github.io/entries-api/",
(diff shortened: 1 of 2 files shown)


tidoust:
The spec mostly contains TODOs so it seemed too early to add it to browser-specs and we put it in the [monitor list](https://github.com/w3c/browser-specs/blob/master/src/data/monitor.json#L11).

That said, this API is a bit special indeed: the `DataCue` interface used to be in HTML but got dropped. It is still in Safari, but was removed from Edge during the move to Chromium. All in all, if it seems useful to have the IDL in WPT already, I'm all fine adding the spec. The PR just needs to be completed to remove the `WICG/datacue` entry from the monitor list.

foolip:
I actually don't want this for WPT, but for https://github.com/foolip/mdn-bcd-collector/issues/917, since it is something shipping in a current browser.

That being said, I'm adding lots of custom IDL to cover that, so I could do that for `DataCue` as well if you'd prefer?

tidoust:
Let's add it to browser-specs then. I completed the PR to remove the entry from the monitor list.

foolip:
Thanks @tidoust!

PR #237, opened 2021-01-25
Update URL of CSS Cascading Level 5 / Monitor Web Audio API v2 repo

### LINKED ISSUE #234: New specs for review

- [ ] https://webaudio.github.io/web-audio-api-v2/ from [WebAudio/web-audio-api-v2](https://github.com/WebAudio/web-audio-api-v2)
- [ ] https://www.w3.org/TR/css-cascade-5/ from [w3c/csswg-drafts](https://github.com/w3c/csswg-drafts)

tidoust:
CSS Cascading Level 5 was published as FPWD last week, so URL needs to be updated.

@deniak, I thought the CSS WG wanted links to the editor's draft to link to the right version, but published versions of levels 4 and 5 link to the unversioned https://drafts.csswg.org/css-cascade/. I would rather have expected:
- https://www.w3.org/TR/css-cascade-4/ to target https://drafts.csswg.org/css-cascade-4/
- https://www.w3.org/TR/css-cascade-5/ to target https://drafts.csswg.org/css-cascade-5/

Also, the versionless https://www.w3.org/TR/css-cascade/ targets Level 5. I would have expected it to continue to target Level 4, which is still a Working Draft. That would be consistent with https://drafts.csswg.org/css-cascade/. Or is that intended?

(Beware if you make updates to the W3C database ;))

This update also adds the the WebAudio API v2 to the monitor list.

Fix #234.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -11,10 +11,6 @@
   "https://drafts.css-houdini.org/font-metrics-api-1/",
   "https://drafts.csswg.org/css-animations-2/ delta",
   "https://drafts.csswg.org/css-backgrounds-4/ delta",
-  {
-    "url": "https://drafts.csswg.org/css-cascade-5/",
-    "shortTitle": "CSS Cascading 5"
-  },
   "https://drafts.csswg.org/css-env-1/",
   "https://drafts.csswg.org/css-extensions-1/",
   {
@@ -296,6 +292,10 @@
   {
     "url": "https://www.w3.org/TR/css-cascade-4/",
     "shortTitle": "CSS Cascading 4"
+  },
+  {
+    "url": "https://www.w3.org/TR/css-cascade-5/",
+    "shortTitle": "CSS Cascading 5"
   },
   "https://www.w3.org/TR/css-color-4/",
   "https://www.w3.org/TR/css-color-5/ delta",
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):

deniak:
> CSS Cascading Level 5 was published as FPWD last week, so URL needs to be updated.
> 
> @deniak, I thought the CSS WG wanted links to the editor's draft to link to the right version, but published versions of levels 4 and 5 link to the unversioned https://drafts.csswg.org/css-cascade/. I would rather have expected:
>     * https://www.w3.org/TR/css-cascade-4/ to target https://drafts.csswg.org/css-cascade-4/
>     * https://www.w3.org/TR/css-cascade-5/ to target https://drafts.csswg.org/css-cascade-5/
> 

As of today, we don't have any requirements regarding the ED for specifications with different levels but in the particular case of `css-cascade`, I guess you are right. @svgeesus can you confirm the ED should be updated? If so, can you also make sure the bikeshed sources are updated with the right ED.

> 
> Also, the versionless https://www.w3.org/TR/css-cascade/ targets Level 5. I would have expected it to continue to target Level 4, which is still a Working Draft. That would be consistent with https://drafts.csswg.org/css-cascade/. Or is that intended?
> 

Yes, correct. The shortlink should only be updated if the WG requests to switch the version. I reverted the change.
/cc @jennyliang220 .

> (Beware if you make updates to the W3C database ;))

I'm no longer allowing myself to touch it :)

PR #252, opened 2021-03-03
Rename CORS RFC1918 to Private Network Access

### LINKED ISSUE #250: New specs for review

- [ ] https://drafts.csswg.org/css-contain-3/ from [w3c/csswg-drafts](https://github.com/w3c/csswg-drafts)
- [ ] https://wicg.github.io/private-network-access/ from [WICG/private-network-access](https://github.com/WICG/private-network-access)

tidoust:
Also adds CSS Containment 3 to monitor list. Spec is flagged as an Exploratory Working Draft and mostly empty for now.

Fixes #250.


dontcallmedom (approved):

foolip:
After thus change and https://github.com/w3c/webref/commit/797d4c5d30a0c6c09fc374d5a3e72fe445cabf94 there's both `cors-rfc1918.idl` and `private-network-access.idl` in webref, which causes the tests to fail. What is the process for deleting the original after a rename?

foolip:
Thanks @tidoust! If we get around to defining a weekly release process for `@webref/idl`, it would be great to have that cleanup job run on the morning of that day.

tidoust:
> I'm re-running a crawl on /TR to fix current errors in the tr folder and will run the cleanup job manually after that to get rid of the now useless files.

Done, original files are now gone.


> it would be great to have that cleanup job run on the morning of that day.

Yup, that's already the case! The cleanup job runs at 2am UTC every Monday.

PR #261, opened 2021-03-15
Add Mike as reviewer

dontcallmedom:
Needed in template too apparently?

### DIFF
--- a/.github/candidate-specs.md
+++ b/.github/candidate-specs.md
@@ -1,6 +1,6 @@
 ---
 title: New specs for review
-assignees: tidoust, dontcallmedom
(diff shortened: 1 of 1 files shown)


tidoust:
> Needed in template too apparently?

Right, previous PR only added Mike to the list of reviewers for pull requests, but not for issues. If you'd like to do this for issues too, why don't you also update `incorrect-base-url.md`?

dontcallmedom:
let's start with the stuff where we needs broader support - all-your-incorrect-base-are-belong-to-urls feels more edge-case-y and not nearly as frequent

PR #262, opened 2021-03-15
Add newly identified specs

### LINKED ISSUE #260: New specs for review

- [ ] https://tc39.es/proposal-error-cause/ from [tc39/proposal-error-cause](https://github.com/tc39/proposal-error-cause)
- [ ] https://tc39.es/proposal-temporal/ from [tc39/proposal-temporal](https://github.com/tc39/proposal-temporal)
- [ ] https://w3c.github.io/device-posture/ from [w3c/device-posture](https://github.com/w3c/device-posture)
- [ ] https://www.w3.org/TR/IndexedDB-3/ from [w3c/IndexedDB](https://github.com/w3c/IndexedDB)

dontcallmedom:
2 new TC39 stage 3 proposals (error cause & temporal); device-posture is a derivative of already included screenfold; IndexDB3 is set to superseded IndexDb-2
fixes #260

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -110,7 +110,6 @@
   "https://url.spec.whatwg.org/",
   "https://w3c.github.io/badging/",
   "https://w3c.github.io/contentEditable/",
-  "https://w3c.github.io/device-posture/",
   "https://w3c.github.io/gamepad/extensions.html",
   "https://w3c.github.io/mathml-aam/",
   "https://w3c.github.io/media-playback-quality/",
--- a/specs.json
+++ b/specs.json
@@ -97,6 +97,7 @@
   },
   "https://tc39.es/proposal-atomics-wait-async/",
   "https://tc39.es/proposal-class-fields/",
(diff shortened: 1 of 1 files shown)


tidoust (changes requested):
Device Posture is not a derivative of Screen Fold, it is the new name of the spec, and https://w3c.github.io/screen-fold/ redirects to https://w3c.github.io/device-posture/

So Screen Fold needs to be removed from the list.

I believe that Device Posture is about to be published to /TR, so we can perhaps wait a couple of days until it's done to update the list, so that we can add the right URL directly.

dontcallmedom:
ha, thanks I had missed that! what change to the PR are you suggesting? leaving screen-fold as is?

tidoust:
> a, thanks I had missed that! what change to the PR are you suggesting? leaving screen-fold as is?

Yes, I would leave screen-fold in the list and ignore device-posture until it gets published to /TR. At worst, we'll get a reminder next week.

tidoust (approved):

PR #289, opened 2021-05-03
Reflect publication of WebAuthn Levle 3

### LINKED ISSUE #287: New specs for review

- [ ] https://www.w3.org/TR/webauthn-3/ from [w3c/webauthn](https://github.com/w3c/webauthn)

dontcallmedom:
close #287

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -612,7 +612,7 @@
   "https://www.w3.org/TR/web-share/",
   "https://www.w3.org/TR/webaudio/",
   {
-    "url": "https://www.w3.org/TR/webauthn-333/",
+    "url": "https://www.w3.org/TR/webauthn-3/",
     "shortTitle": "Web Authentication"
   },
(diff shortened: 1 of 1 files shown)


specs.json
```diff
@@ -612,7 +612,7 @@
   "https://www.w3.org/TR/web-share/",
   "https://www.w3.org/TR/webaudio/",
   {
-    "url": "https://www.w3.org/TR/webauthn-2/",
+    "url": "https://www.w3.org/TR/webauthn-333/",
```
tidoust, line 615: One `3` might be enough...

```suggestion
    "url": "https://www.w3.org/TR/webauthn-3/",
```

tidoust (approved):

specs.json
dontcallmedom: that's called being ahead of the game!

PR #290, opened 2021-05-04
Add Web Animations 2 and CSS Fonts 5

tidoust:
Follows from discussions in:
https://github.com/w3c/browser-specs/issues/279#issuecomment-831634786

The specs are flagged with a "not ready for implementation" banner, but already documented in BCD in practice, and implemented in at least one browser.


dontcallmedom (approved):
Chromium just announced its intent to ship for `size-adjust`, further confirming implementations are already under way https://groups.google.com/a/chromium.org/g/blink-dev/c/1PVr94hZHjU/m/J0xT8-rlAQAJ

tidoust:
This suggests that it might indeed be time to drop the banner...

PR #299, opened 2021-05-24
Switch ecma262 spec to multipage version

sideshowbarker:
There’s now a multipage version of the ECMAScript spec, at https://tc39.es/ecma262/multipage/. This change switches to the multipage version, rather than the single-page version. cc @Elchi3

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -65,7 +65,8 @@
   "https://streams.spec.whatwg.org/",
   "https://svgwg.org/specs/animations/",
   {
-    "url": "https://tc39.es/ecma262/",
+    "url": "https://tc39.es/ecma262/multipage/",
+    "multipage": true,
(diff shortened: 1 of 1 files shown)


sideshowbarker:
The current need for this is https://github.com/mdn/browser-compat-data/pull/10615, which is a patch to switch all the BCD spec URLs for the ES spec over to the multipage spec.

tidoust:
The makes sense to me but this won't fully give you what you want as long as Specref is not updated as well, because browser-specs fetches the nightly URL and other spec info from Specref. I created tobie/specref#661 accordingly.

A side benefit of updating Specref is that references to ECMAScript from other specs will also swtich to the multipage version, which seems a good idea.

tidoust (approved):
URL already updated in Specref, so we're good to go!

PR #318, opened 2021-06-21
Add Web Neural Network API

### LINKED ISSUE #317: New specs for review

- [ ] https://webmachinelearning.github.io/webnn/ from [webmachinelearning/webnn](https://github.com/webmachinelearning/webnn)

tidoust:
Fixes #317.


tidoust:
We'll have to update the URL when the spec gets published as FPWD (which should be very soon) but that won't change the shortname.

dontcallmedom (approved):

PR #324, opened 2021-07-01
Make "main" branch the default branch

tidoust:
Prepare migration to `main` as default branch.


dontcallmedom (commented):
the `master` branch name is also used in the $id$ of the various JSON schemas

tidoust:
> 
> 
> the `master` branch name is also used in the $id$ of the various JSON schemas

Ah. I had a large file in the root folder locally and the search in files feature I'm using basically stops searching with a "file too large, not going any further" message which I had of course completely missed. Interesting feature :)

dontcallmedom (approved):

Issue #50, opened 2020-05-26
Consider scanning for missing specs


foolip:
In a [check-missing-specs.js](https://github.com/foolip/day-to-day/blob/e894f3c28e8e8f538a36a9c909f515e758311e2d/build/check-missing-specs.js) script for another project I wrote some code to discover specs missing from my list. I'm going to abandon that, but some of these tricks might be useful for this project:
- Look for all repos in https://github.com/w3c and https://github.com/WICG
- Load `https://${org}.github.io/${repo.name}/` for each one
- If they load, consider them a candidate

However, assuming that https://github.com/w3c/validate-repos is being run regularly, then this alternative may be better:
- For each repo check `repo-type` in `w3c.json` for the spec-like types

My code didn't include it, but for WHATWG we now have https://github.com/whatwg/sg/blob/master/db.json as a centralized source of information, which could also be used to look for what's missing in w3c/browser-specs.

tabatkins:
Related: I just did a diff between Shepherd's set of known spec urls and the urls from here, and got the following results:

```
92 Shepherd URLs not in W3C data:
* http://www.w3.org/TR/mixed-content/
* http://www.w3.org/TR/referrer-policy/
* https://dev.w3.org/webfonts/WOFF2/spec/
* https://drafts.css-houdini.org/box-tree-api-1/
* https://drafts.css-houdini.org/css-parser-api-1/
* https://drafts.css-houdini.org/scroll-customization-api-1/
* https://drafts.csswg.org/css-2007/
* https://drafts.csswg.org/css-2010/
* https://drafts.csswg.org/css-2015/
* https://drafts.csswg.org/css-2017/
* https://drafts.csswg.org/css-2018/
* https://drafts.csswg.org/css-2020/
* https://drafts.csswg.org/css-box-4/
* https://drafts.csswg.org/css-egg-1/
* https://drafts.csswg.org/css-mobile/
* https://drafts.csswg.org/css-page-template-1/
* https://drafts.csswg.org/css-preslev-1/
* https://drafts.csswg.org/css-print/
* https://drafts.csswg.org/css-shadow-parts-1/
* https://drafts.csswg.org/css-speech-1/
* https://drafts.csswg.org/css-template-1/
* https://drafts.csswg.org/css-tv-1/
* https://drafts.csswg.org/css-viewport-1/
* https://drafts.csswg.org/mediaqueries-3/
* https://drafts.fxtf.org/custom-1/
[...]

PR #326, opened 2021-07-01
Add fingerprinting-guidance to the list of specs.

jyasskin:


dontcallmedom (changes requested):
hi @jyasskin - the Fingerprinting Guidance document doesn't a prori fulfills the [requirements for addition here](https://github.com/w3c/browser-specs#spec-selection-criteria), since it's not a spec that browsers would implement. Can you say more about the motivation of having added?

Thanks!

tidoust (approved):
We did not include the spec in the initial list because it is a Note that is not "implemented" per se and we hadn't fleshed out [spec selection criteria](https://github.com/w3c/browser-specs/#spec-selection-criteria) at the time (see #26 and https://github.com/w3c/browser-specs/issues/50#issuecomment-645489318) .

These criteria now mention "guidelines from horizontal activities such as accessibility, internationalization, privacy and security" as a good motivation for inclusion, and it makes sense for specs to reference the browser fingerprinting spec. So +1 from me.

@jyasskin FYI, I note that the spec only exports 2 definitions for the time being:
- [browser fingerprinting](https://w3c.github.io/fingerprinting-guidance/#dfn-browser-fingerprinting)
- [fingerprinting surface](https://w3c.github.io/fingerprinting-guidance/#dfn-fingerprinting-surface)

There could be other terms worth exporting so that other specs may cross-reference them such as [passive fingerprinting](https://w3c.github.io/fingerprinting-guidance/#dfn-passive-fingerprinting), [active fingerprinting](https://w3c.github.io/fingerprinting-guidance/#dfn-active-fingerprinting) and [cookie-like fingerprinting](https://w3c.github.io/fingerprinting-guidance/#dfn-cookie-like-fingerprinting).

It may also be useful to be add the best practices themselves to the xref database, I'm not sure there is a good mechanism for that for the time being though.

dontcallmedom (approved):

tidoust:
@dontcallmedom I note that I hadn't seen your feedback before I provided mine. If you believe that we need to give it more thoughts, that's fine with me. Sooner or later, we'll hit the "spec not for browsers but needs to be in the list coz' that other spec needs to reference terms" problem 🙀

dontcallmedom:
> @dontcallmedom I note that I hadn't seen your feedback before I provided mine. If you believe that we need to give it more thoughts, that's fine with me. Sooner or later, we'll hit the "spec not for browsers but needs to be in the list coz' that other spec needs to reference terms" problem scream_cat

That was indeed kind of what I were I was going - if indeed the need is/was mostly in terms of expanding the set of available definitions, this might be sign that time has come to bite that bullet. But I also recognize your argument about the updated set of acceptable documents applies, so :shrug:

jyasskin:
Sorry for taking a while to reply: my reasoning was what @tidoust [described](https://github.com/w3c/browser-specs/pull/326#pullrequestreview-698006462), that other implemented specs need to refer to the couple exported definitions. The other plausible terms aren't exported yet just because other specs happen not to have needed to refer to them. I don't edit that document, but I suspect its editor will accept such pull requests when they're needed. I think we'd add the best practices themselves similarly to how https://w3ctag.github.io/ethical-web-principles/#oneweb does it: just make the title a `dfn`.

PR #329, opened 2021-07-02
Use latest version of github-push-action

tidoust:
Build workflow used an old version of github-push-action, which used `master`as default branch name. The latest version uses the actual default branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,6 @@ jobs:
         git config user.email "<>"
         git commit -m "[data] Update spec info" -a || true
     - name: Push changes
-      uses: ad-m/github-push-action@master
+      uses: ad-m/github-push-action@0.6.0
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -27,6 +27,6 @@ jobs:
         git config user.email "<>"
         git commit -m "[data] Update spec info" -a || true
     - name: Push changes
-      uses: ad-m/github-push-action@v0.5.0
+      uses: ad-m/github-push-action@master
```
dontcallmedom, line 30:
```suggestion
      uses: ad-m/github-push-action@0.6.0
```
it feels risky to automatically adopt upgrades to github actions with access to privileges rights without at least some review

dontcallmedom (approved):

tidoust:
Argh, missed the "v" :( I'll push the fix to default branch directly.

tidoust:
Well, and use of the default branch name only exists in latest version, not in `v.0.6.0`... I'll force the name of the branch as parameter.

PR #332, opened 2021-07-02
Update documentation re repository.nightly availability

### LINKED ISSUE #331: IETF specs provides no `nightly.repository`
It's contradictory to https://github.com/w3c/browser-specs#nightlyrepository.

>The `repository` property is always set.

dontcallmedom:
close #331
Follow up to https://github.com/w3c/browser-specs/pull/325/commits/687e149c501167e2a189c28da28f648f80717fe6

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,7 +377,7 @@ That path is computed by parsing the contents of the repository for common
 patterns. The info must be specified in `specs.json` for specifications that do
 not follow a common pattern.
 
-The `sourcePath` property is always set.
+The `sourcePath` property is always set when `repository` is set.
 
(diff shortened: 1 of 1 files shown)


dontcallmedom:
Since we broke a commitment, it does feel like this should have been a major release…

dontcallmedom:
(I thought we didn't have such a commitment since the schema didn't enforce it - maybe another worth looking at more generally)

saschanaz:
Note that `sourcePath` is now also undefined-able too

tidoust (approved):
> Since we broke a commitment, it does feel like this should have been a major release…

True. That's a bit too late but I'll issue a new major release.

PR #349, opened 2021-07-12
Switch RFC URLs to rfc-editor.org

### LINKED ISSUE #342: Use rfc-editor.org for IETF specs 8650+
As [pointed out in the discussion on adding IETF specs](https://github.com/w3c/browser-specs/issues/280#issuecomment-872379335), and [starting from RFC8650](https://www.rfc-editor.org/format-live/), IETF RFCs get a much nicer HTML rendering on `rfc-editor.org` than they get on `datatracker.ietf.org`.

We should thus point toward these URLs (at least via nightly as we currently do for httpwg.org URLs?) at least for these RFCs - but maybe more generally.

This is probably a discussion that should start upstream in specref.

dontcallmedom:
close #342

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -6,36 +6,6 @@
     }
   },
   "https://console.spec.whatwg.org/",
-  {
-    "url": "https://datatracker.ietf.org/doc/html/rfc2397",
-    "groups": [
-      {
-        "name": "Network Working Group",
-        "url": "https://datatracker.ietf.org/group/app/"
-      }
-    ]
-  },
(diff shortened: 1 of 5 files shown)


tidoust (changes requested):
Tests on the generated index file with IETF specific logic should be updated with rfc-editor URLs as well:
https://github.com/w3c/browser-specs/blob/main/test/index.js#L105

tidoust (approved):

PR #352, opened 2021-07-19
Monitor MathML 4 spec

tidoust:
I don't really know what to do with this spec, raised in #350. I propose to add it to the monitor list for now. Spec is under development within the MathML WG but it is mostly a copy of MathML 3 for now, is not yet based on MathML Core and, as opposed to MathML Core that specifically targets browsers, is likely going to define features that won't be implemented by browsers.

### DIFF
--- a/src/data/ignore.json
+++ b/src/data/ignore.json
@@ -41,6 +41,9 @@
     }
   },
   "repos": {
+    "w3c/mathml": {
+      "comment": "not targeted at browsers"
+    },
     "w3c/adpt": {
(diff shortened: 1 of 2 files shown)


dontcallmedom:
given that it's unlikely to be browser targeting, is there any reason not to put it in the ignore list?

tidoust:
I moved it to the ignore list. It's the "unlikely" that made me err on the monitor side, but that should be "_very_ unlikely" in practice.

Issue #238, opened 2021-01-25
Check no duplicate in browser and ignore/monitor lists


tidoust:
There are already tests in place to make sure that an entry in the ignore list does not overlap with an entry in the monitor list. We should also make sure that there are no overlap between the ignore/monitor lists and the final browser list.

Ideally, these tests would directly detect cases when we add an entry to the browser list but forget to remove it from the monitor list (as in #233). Note that this means that the test needs to process `specs.json`, and not `index.json` since the updated version won't appear in the PR. If testing against `specs.json` turns out to be a pain, an alternative is to check this periodically and create PRs as needed.

PR #354, opened 2021-07-23
Add preferCurrentTab spec

foolip:


tidoust (changes requested):
@dontcallmedom put it on the monitor list early July:
https://github.com/w3c/browser-specs/blob/main/src/data/monitor.json#L3

If implementation plans are already clear, fine adding it to the list. The entry needs to be removed from the monitor list though (FWIW, automatic detection that this is needed is tracked in #238)

foolip:
It may soon ship in Chromium:
https://groups.google.com/a/chromium.org/g/blink-dev/c/YoefXLTQsw0/m/NVoV6NcIAgAJ

PR #377, opened 2021-09-13
Add source path for Scripting Policy spec

tidoust:
The code cannot find the source path by itself.


dontcallmedom (approved):

tidoust:
FWIW, I believe the tests fail because they run against the current `index.json` file which has the problem that this PR is going to fix.

PR #378, opened 2021-09-16
feat: add https://tc39.es/proposal-intl-locale-info

romulocintra:
Add Proposal : https://tc39.es/proposal-intl-locale-info

This fix's https://github.com/mdn/browser-compat-data/pull/12314

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -92,6 +92,7 @@
   "https://tc39.es/proposal-class-static-block/",
   "https://tc39.es/proposal-error-cause/",
   "https://tc39.es/proposal-import-assertions/",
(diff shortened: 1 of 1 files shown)


romulocintra:
@dontcallmedom  the #379 affects this PR to land? There is something else that needs to be done?

tidoust (approved):
Not at all, @romulocintra! PR is good to merge. It just made us realize that we were not taking Intl proposals into account when looking for new specs to add, whereas we should.

As a matter of fact, there are a few other Intl proposals that should get added too. We'll do that afterwards.

Thanks for preparing the PR!

PR #389, opened 2021-09-27
Monitor Intl.NumberFormat, fix shortnames of Intl.DisplayNames

### LINKED ISSUE #388: Improve series shortname for versioned TC39 proposals
Code currently generates:
- `intl-displaynames-v` for https://tc39.es/intl-displaynames-v2/
- `tc39-intl-numberformat-v` for https://tc39.es/proposal-intl-numberformat-v3/

The `-v` should be dropped.

tidoust:
The Intl.NumberFormat proposal may be at stage 3 but it only exists as a README for now (and README claims it is at stage 2...). Our code cannot set the `sourcePath` as a result and it seems premature to add the spec as a result.

Also, the `tc39-` prefix was missing from the Intl.DisplayNames shortname and the series shortname didn't look too good. Code could probably be fixed to support `-vXX` URL suffixes, but then there is only one occurrence of such a URL for the time being, so handling it directly in specs.json.

Close #388.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -87,7 +87,10 @@
   },
   {
     "url": "https://tc39.es/intl-displaynames-v2/",
-    "shortname": "intl-displaynames-v2"
+    "shortname": "tc39-intl-displaynames-v2",
+    "series": {
+      "shortname": "tc39-intl-displaynames"
+    }
   },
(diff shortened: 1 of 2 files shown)


tidoust:
Note test fails on the previously generated `index.json`, that this PR would fix.

dontcallmedom (approved):

PR #394, opened 2021-10-04
Remove frame-timing, as no longer being pursued

yoavweiss:


tidoust:
Thanks! I completed the PR to add the spec to the ignore list as well (otherwise the action that reports on new specs would likely report it as a new spec next time it runs).

I note it would be even better if the spec repo and the spec itself were updated to note that it is no longer being pursued!

PR #396, opened 2021-10-05
Use new URL for WebIDL spec as WHATWG living standard

dontcallmedom:


tidoust (approved):
Sounds good. We'll need to tweak Reffy before integrating this, as it has the URL of WebIDL hardcoded somewhere, IIRC.

PR #413, opened 2021-10-15
Remove mediacapture-transform as it is a non-adopted proposal

jan-ivar:
See https://lists.w3.org/Archives/Public/public-webrtc/2021Oct/0025.html

cc @dontcallmedom

### DIFF
--- a/src/data/monitor.json
+++ b/src/data/monitor.json
@@ -44,10 +44,6 @@
       "comment": "unofficial draft",
       "lastreviewed": "2021-09-01"
     },
-    "w3c/mediacapture-transform": {
-      "comment": "still marked as unofficial draft",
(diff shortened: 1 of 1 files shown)


dontcallmedom:
@jan-ivar the monitor list is for repos that precisely aren't considered part of "browser-specs" yet; the repo is temporarily redirected, but given that it may re-appear to point to the possible options in this space, I'm not sure it makes sense to remove it from the monitor list.

jan-ivar:
@dontcallmedom Makes sense. I see this is explained in [candidate-specs.md](https://github.com/w3c/browser-specs/blob/main/.github/candidate-specs.md) (assuming `monitor-repo.js` is `monitor.js`).

PR #421, opened 2021-11-01
NTLM over HTTP spec

hamishwillee:
This adds NTLM spec https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-ntht/ 

I am not sure if this is the "right" way to do this / link to the published version.

This is to allow linking to NTLM spec from MDN browser compatibility data: https://github.com/mdn/browser-compat-data/pull/12446

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -6,6 +6,7 @@
     }
   },
   "https://console.spec.whatwg.org/",
+  "https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-ntht/",
   "https://dom.spec.whatwg.org/",
   "https://drafts.css-houdini.org/css-typed-om-2/ delta",
   "https://drafts.css-houdini.org/font-metrics-api-1/",


tidoust:
This is not going to work as-is because there's nothing that the code can infer from such a URL. Typically, the code does not know how to create a suitable shortname for the spec (probably should be `ms-ntht`).

That said, I'd be tempted to say that this spec is out of scope for browser-specs. One of the [spec selection criteria](https://github.com/w3c/browser-specs#spec-selection-criteria) is that the spec should be "developed by a well-known standardization or pre-standardization group". If I understand things correctly, this spec is developed and maintained by a single vendor.

A couple of other considerations that could no doubt be addressed if needed:
- Another selection criterion that I realize is not expressed anywhere, is that it should be possible to process the spec somehow with automated tools. This is harder to achieve here since the spec does not follow any of the formatting rules that others do. At least, we don't have code in browser-specs or Reffy that can process a PDF document.
- We try to target specs themselves in browser-specs. In this case, that would probably mean the [PDF version](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-NTHT/%5bMS-NTHT%5d.pdf). At the same time, it's not necessarily a good idea to link to a PDF directly.

@dontcallmedom, WDYT?

dontcallmedom:
I agree there would be limited value in adding it to browser-specs at the moment.

@hamishwillee I suggest instead documenting it as an exception to the [BCD url test](https://github.com/mdn/browser-compat-data/blob/3985c3e0b628b542997b9b5a6ac0123aca6f2181/test/spec-urls.test.js#L27)

hamishwillee:
Thanks very much for the information and the guidance. Absolutely happy with this answer - I suspected it might be something like this, but new to this repo.

Issue #360, opened 2021-08-06
Inconsistent titles pre-FPWD


dontcallmedom:
The [nightly of WAI ARIA](https://w3c.github.io/aria/) is numbered 1.3, but browser-spec associates it with a 1.2 title, presumably because that's what specref knows about since it hasn't reached FPWD.

I wonder if there is a systematic fix to this situation that browser-specs should consider.

(raised from https://github.com/mdn/content/issues/7651#issuecomment-894137711)

tidoust:
We could set the title from the nightly version instead of from the release version, but then there would be no way to retrieve the title of the WAI-ARIA 1.2 spec, which seems strange considering that the WAI-ARIA entry in browser-specs uses the /TR URL of version 1.2 as root `url` (`https://www.w3.org/TR/wai-aria-1.2/`) and `wai-aria-1.2` as `shortname`.

A better fix might be to introduce a `title` property at the `nightly` level.

FWIW, the title actually comes from the W3C API, not from Specref (Specref has the same info in any case since it also uses the W3C API as source)

Issue #359, opened 2021-08-06
Provide version-less titles for series


dontcallmedom:
When a consumer (such as BCD/MDN) is not interested in a specific version of a spec, but mostly its latest in the series, there is no good way to refer to the work since the only thing the data provides is the title of the latest version.

Maybe series should have a title - for W3C specs, the series end point of the W3C API provides such a title.

dontcallmedom:
(raised from https://github.com/mdn/content/issues/7651#issuecomment-894091069)

tidoust:
> Maybe series should have a title - for W3C specs, the series end point of the W3C API provides such a title.

I don't see where the W3C provides a version-less title at the series level (or any kind of title at the series level, actually), could you provide an example?

Also, what would the version-less title look like for say, the wai-aria series or the css-color series?

dontcallmedom:
> I don't see where the W3C API provides a version-less title at the series level (or any kind of title at the series level, actually), could you provide an example?

See e.g. https://api.w3.org/specification-series/wai-aria - it has a `name` property set to "Accessible Rich Internet Applications (WAI-ARIA)"

tidoust:
Ah, I was looking for a `title` property as everywhere else. That would have been too simple ;)

I'm not sure whether that series title is automatically generated. For instance, it returns "CSS Color" for the css-color series, and a naive approach would rather generate "CSS Color Module". It could be interesting to align this with the code that we have in place to generate the `shortTitle` property.

Anyway, we could leverage the W3C API whenever possible to create the version-less `series.title` property, and an extended version of the `shortTitle` code that also drops the version number when the series is not known to the W3C API.

PR #447, opened 2021-12-10
Add RFC6265bis draft

### LINKED ISSUE #345: Dealing with experimental IETF RFC
Hi!

I've hit an edge case that sounds legit.

The official spec is RFC 6265 published in 2011 (and listed here).
Since 2017, work is underway to publish an extension https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis.

Every six months or so, a new version is published: https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis-08 is the latest, with a validity until the end of the year.

There are widely implemented features available only in the draft (E.g. [Set-Cookie/SameSite on MDN Web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite)).

How should we deal with this case? Add https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis as a nightly entry to RFC6265?

dontcallmedom:
close #345

I've tested the built index with reffy which seemed to happily extract references and links

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -6,6 +6,22 @@
     }
   },
   "https://console.spec.whatwg.org/",
+  {
+    "url": "https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-rfc6265bis",
+    "shortname": "rfc6265bis",
+    "organization": "IETF",
+    "groups": [
+      {
+        "name": "HTTP Working Group",
+        "url": "https://datatracker.ietf.org/wg/httpbis/"
+      }
+    ],
+    "nightly": {
+      "url": "https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html",
+      "repository": "https://github.com/httpwg/http-extensions",
+      "sourcePath": "draft-ietf-httpbis-rfc6265bis.md"
+    }
+  },
   "https://dom.spec.whatwg.org/",
   "https://drafts.css-houdini.org/css-typed-om-2/ delta",
   "https://drafts.css-houdini.org/font-metrics-api-1/",


tidoust (commented):
Shouldn't the spec also be defined with:

```
{
  "series": {
    "shortname": "rfc6265"
  }
}
```

... so that it gets associated with RFC6265, which it obsoletes? That may require some adjustments to the logic that sets `seriesPrevious` and `seriesNext` though.

Regarding the title, have you checked how many entries end up with a different title with the new logic? I agree that the new logic is as bad as the previous one; the change may just come as a surprise to people already using browser-specs and who may wonder why e.g. the version number in the title of a spec is no longer the one they were used to seeing.

(All in all, I think that we should deprecate the `title` property at the root level and rather have one at the `series`, `release` and `nightly` levels, which is more or less where we're heading with #359 and #360)

dontcallmedom:
> Shouldn't the spec also be defined with:
> ```
> {
>   "series": {
>     "shortname": "rfc6265"
>   }
> }
> ```

Maybe, but
* What level would I set it to? (I'm assuming I can't just use "bis" as a level
* whenever it actually replaces it, the series will change shortname, which kind of defeats the purpose of using the shortname
 

> Regarding the title, have you checked how many entries end up with a different title with the new logic? I agree that the new logic is as bad as the previous one; the change may just come as a surprise to people already using browser-specs and who may wonder why e.g. the version number in the title of a spec is no longer the one they were used to seeing.

As far as I can tell, none:
```sh
$ git diff -- index.json|grep title
+    "title": "Cookies: HTTP State Management Mechanism",
```
 
> (All in all, I think that we should deprecate the `title` property at the root level and rather have one at the `series`, `release` and `nightly` levels, which is more or less where we're heading with #359 and #360)

+1

tidoust (approved):
> whenever it actually replaces it, the series will change shortname, which kind of defeats the purpose of using the shortname

Which makes me wonder whether we actually need the two specs in browser-specs. We could perhaps just have RFC6265bis (with an `rfc6265` shortname). That said, I don't think that having the two specs creates a confusion in the tools we have, so fine keeping the specs separate for now.

> As far as I can tell, none:

Which seemed weird so I double checked. In practice, that's because the update does not do what its commit message suggests it does ;) The update only fetches the title from the nightly URL when:
1. the spec is not known in the W3C API and Specref
2. the entry in `specs.json` explicitly sets a nightly URL

There is only one entry in the list that matches these criteria and that's RFC6265bis. I guess we could just as well have stuck with a hardcoded title in `specs.json`. Anyway, the update is fine.

dontcallmedom:
thanks - I agree my commit message was a bit misleadingly generic :)

PR #487, opened 2022-02-01
Remove mediacapture depth spec

dontcallmedom:
discontinued by the WG

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -623,7 +623,6 @@
       "sourcePath": "media-source-respec.html"
     }
   },
-  "https://www.w3.org/TR/mediacapture-depth/",
   "https://www.w3.org/TR/mediacapture-fromelement/",
   "https://www.w3.org/TR/mediacapture-streams/",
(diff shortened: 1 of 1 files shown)


tidoust (commented):
Shouldn't the w3c.json file be updated in the mediacapture depth repo too so that our `find-specs.js` tool does not prompt us to add the spec again?

If not, the spec needs to be added to the ignore list.

dontcallmedom:
well, technically, it's still on the Rec-track (since the Discontinued Draft status is a stage there).

I think we need to update find-spec to ignore discontinued drafts; we may also consider having a clean-up script that alert us if browser-spec includes discontinued drafts?

dontcallmedom:
actually, with the repo archived it won't appear in the validate-repo report; I need to update spec-dashboard to make it ignore Discontinued drafts, but otherwise, I think it would already be ignored without further ado.

tidoust (approved):

PR #491, opened 2022-02-02
Reflect new URLs of event-timing and largest-contentful-paint

dontcallmedom:
They've moved from WICG

I filed https://github.com/WICG/wicg.github.io/issues/16 since the old URLs are now 404


tidoust (approved):
That's good but, on top of redirects, the specs also needed to be dropped from WICG's biblio file (done in https://github.com/WICG/admin/pull/149), and Specref needs to be updated (done in https://github.com/tobie/specref/pull/695).

tidoust:
> Should we bring this up on WICG/admin to streamline this (and more importantly, not making us be the ones ending up doing the necessary admin work)?

Having a checklist in place couldn't hurt.

Issue #436, opened 2021-11-22
Support for non-browser specs


tidoust:
This issue tracks the extension of the browser-specs to cover non-browser specs.

The browser-specs was created to track specs that target Web browsers. Once in a while, selection criteria are a bit fuzzy but list boundaries have remained relatively clear so far and have shown value for integration in tools and projects such as Webref, MDN, WPT.

Now, some tools actually need to ingest more than just "browser" specs. Notably, spec authoring tools would happily extend their cross-referencing (xref) terms database so that other specs may benefit from it as well. For instance, a given WoT spec may need to reference a WoT term defined in another WoT spec. Currently, it cannot rely on the xref mechanism for that.

We already had to add a few specs to browser-specs that don't meet selection criteria, e.g. the Process and Patent Policy specs. This somewhat works but can break expectations from tools that rely on browser-specs to stick to actual browser specs, see https://github.com/w3c/browser-specs/issues/390#issuecomment-973984708

Introducing other specs may create other side effects. For instance, the WoT Scripting API defines WebIDL that is not implemented in browsers and should not be included in the `@webref/idl` package.

It seems useful to have a way to open up browser-specs to non-browser specs. It should equally be easy to get back to a browser-only list of specs. We could perhaps maintain two lists of specs in the browser-specs repo. Or perhaps keep one list and flag non-browser specs somehow.

tidoust:
Reopening as "fixes" reference in merged PR was wrong...

foolip:
Could such specs go in a separate package, and ideally a separate repo?

tidoust:
That is one option. We still don't really have a clear view of what we want to end up with in practice and it affects more than just the browser-specs repo. There will be scenarios where people will want to merge the lists and scenarios they will want to consider lists separately. In the xref example, goal is probably to end up with only one database, currently at Webref. In the WoT example, we would want the IDL to be separated from the rest of the IDL.

We may end up duplicating browser-specs and webref, but that introduces a bit of maintenance overhead and will impact some of the tools as well.

foolip:
I think it's a wonderful feature of browser-specs that it only includes things that are implemented in some browser. It would be great if the browser-specs keeps being that, it's in the name after all.

If this repo is renamed all-the-specs and only a subset ends up in the browser-specs package, that would be fine for my usage.

tidoust:
[...]

PR #496, opened 2022-02-07
Add support for non-browser specs

tidoust:
This addresses #436.

This update brings a number of changes to the preparation and release of browser-specs as an NPM package in order to support extending the list of specs to also contain specs that are not targeted at browsers.

A new `categories` property was added to the description of specs. It contains a list of categories that the spec belongs to. The only category so far is `"browser"` for specs that target web browsers. That category is automatically computed based on the name of the group that develops the spec for now. Explicit info in `specs.json` may be used to override the computed value.

The code reuses the list of groups and repos in `src/data/ignore.json` to identify specs that do not target web browsers for the time being. That list should get split into two lists on a short term basis: a list of groups and repos that `find-specs.js` should indeed ignore, and a list of groups and repos linked to specs that are not targeted at browsers (we won't ignore them anymore, but we won't flag them with a `"browser"` category). That split is not done in this update. In general, I expect the logic in `compute-categories.js` to evolve based on non-browser specs that we add to the list.

The release workflow is a copy-and-paste from the one used in webref:
- A new `packages` folder contains the list of packages that can be released. Two packages: `web-specs` which contains all the specs, and `browser-specs` which contains all specs that target Web browsers.
- A `prepare-packages.js` script prepares the contents of these packages.
- A `bump-packages-minor.js` script bumps the minor version of these packages whenever a new spec gets added to the package or when a spec gets removed.
- A `prepare-release.js` script prepares a pre-release PR per package that contains the diff to release (the actual PR is a patch bump on the package version)
- A `release-package.js` script actually releases the packages to NPM when the pre-release PR gets merged.

Jobs were adjusted accordingly. In particular, the build job now only commits updates provided that tests pass, so as not to end up with a committed list that we know is wrong. It also creates the pre-release PRs as needed.

The release job tags the released commit on the main branch, and updates the `web-specs@latest` and `browser-specs@latest` branches. The `web-specs@latest` is the one intended to be published to GitHub Pages.

Notes:
- The packages READMEs are automatically completed based on the main README.
- We will lose the changelog and the creation of Git releases with the new release mechanism, but the changelog was full of opaque "Data updated" in any case, so that does not seem to be a big deal.
- It could be interesting to avoid duplication of common scripts between browser-specs and webref. The scripts are sufficiently similar that this seems worth pursuing, and sufficiently different to make it a bit painful. Plus I'm not too crazy about having to create yet another repo for hosting that common glue code. I propose to address that later on.

Things to do before the PR may be merged:
- Create a first version of the `web-specs` npm package
- Add the npm token secret
- Create the `web-specs@latest` and `browser-specs@latest` branches
- Switch to `web-specs` in Reffy (otherwise we'll lose specs such as the W3C Patent Policy or the i18n glossary from Webref)

Things to do before we add further non-browser specs to the list:
- Add filtering code in Webref curation and release jobs


src/compute-categories.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  let list = [];
+  const requestedCategories = (typeof spec.categories === "string") ?
+    [spec.categories] :
+    (spec.categories || []);
+
+  // All specs target browsers by default unless the spec object says otherwise
+  if (!requestedCategories.includes("reset")) {
+    const browserGroup = !spec.groups.find(group => nonbrowserGroups[group.name]);
```
dontcallmedom, line 40: if I read this correctly, having a non-browser group listed as co-owning a spec disqualifies the spec as a browser spec? if so, while that may be OK today, it feels like this should be documented as potentially wobbly

src/data/ignore.json
```diff
@@ -27,12 +30,18 @@
     "EPUB 3 Working Group": {
       "comment": "specs not targeted at browsers"
     },
+    "Internationalization Working Group": {
```
dontcallmedom, line 33: I don't think that's a broadly true stable statement; I think I would prefer the specs to be labeled as non-browser individually.

dontcallmedom:
kudos again on a giant piece of work!

src/compute-categories.js
tidoust: I had "In general, I expect the logic in compute-categories.js to evolve based on non-browser specs that we add to the list" as a disclaimer in the commit message. I'll add something along these lines to the source code.

In particular, the logic is just meant to work today to tag the 4 specs in the list that don't target browsers, but I agree that we'll need to improve it. I would prefer to do that based on a concrete list of specs, meaning when we actually start adding non-browser specs.

(Typically, I cannot think of an example of a joint deliverable that would qualify as a browser spec but for which one of the groups would fall into the list of non browser groups. Do you have one in mind?)

src/data/ignore.json
tidoust: Which of the specs that the group develop would you qualify as browser specs?
https://www.w3.org/groups/wg/i18n-core/publications

It seems to me that a browser spec is going to be the exception more than the rule.

But, as above, this is currently only meant to work today and likely going to evolve. I'm not even sure that we'll have that many groups that we can flag as "non browser". I'm fine dropping that for now and to rather hardcode the exception in `specs.json`.

src/compute-categories.js
tidoust: Well, I note the code already has a "Logic (and initial spec properties used to compute the list) will likely be adjusted over time" comment near the top. I'll add a more specific comment to that part.

PR #504, opened 2022-02-09
Fix reference for minor bump script

dontcallmedom:
@ is a reserved character in refering to a revision in the git CLI

### DIFF
--- a/src/bump-packages-minor.js
+++ b/src/bump-packages-minor.js
@@ -23,7 +23,7 @@ function specsMatch(s1, s2) {
 function isMinorBumpNeeded(type) {
   // Retrieve contents of last committed index file
   const res = execSync(
-    `git show ${type}@latest:index.json`,
+    `git show ${type}\@latest:index.json`,
     { encoding: 'utf8' }).trim();
   let lastIndexFile = JSON.parse(res);
 


dontcallmedom:
https://github.com/w3c/browser-specs/runs/5128428689?check_suite_focus=true

tidoust:
I meant, where did you get that "@" is a reserved character for Git CLI. I did not find any reference to that in the man pages of Git for instance.

tidoust:
Note that does not seem to fix the problem:
https://github.com/w3c/browser-specs/runs/5129124542?check_suite_focus=true

dontcallmedom:
re @, see https://www.git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emem 
re still failing, the console still shows a \-less output - I'm guessting it needs further escaping due to JS escaping itself

dontcallmedom:
I had written `\-less`, but escaping ate it here too :)

dontcallmedom:
#505 adds another `\` (seems to work locally)

tidoust:
I believe the command rather fails because the checkout action does not "initialize" (for some definition of the term) the `browser-specs@latest` branch. I can typically reproduce that if I:

1. start from a clean `git clone` of the repo
2. run `git show browser-specs@latest:index.json`

If I do a `git checkout browser-specs@latest` first, then the branch appears locally, and the `git show` command then runs successfully.

dontcallmedom:
ha, that may well be true; I had to do that dance for the web-specs@ to get it to work, but thought this was because I hadn't pulled that branch before

tidoust:
Looking into it some more, as long as you don't create a local ref for the `browser-specs@latest` branch, it only exists as `refs/remotes/origin/browser-specs@latest`. Checking it out will create `refs/heads/browser-specs@latest`, which can then be used, abbreviated to `browser-specs@latest`, in the `git show` command.

Possible solutions:
1. Run `git checkout` commands before running `git show` to create the local references. But then we don't really need to checkout the branches.
2. Adjust the `git show` command to target the right refname as in `git show refs/remotes/origin/browser-specs@latest:index.json`. That should work but then I don't like it that this implies that `origin` is always the name of the main origin. Typically, I never have an origin named `origin` in my local repo clones because I prefer more explicit names. We could pass the origin name as parameter to the script, but that seems tedious. Which brings me to...
3. Start with issuing a `git show-ref browser-specs@latest` command to retrieve the ref name, and use it in the `git show` command. If there is more than one possible remote refs, which could happen in local clones of the repo, fail (no way to select the right origin if there is more than one).

PR #505, opened 2022-02-09
Escape \ sign itself

dontcallmedom:


tidoust (approved):
Let's give it a try but I think that the problem is elsewhere, see https://github.com/w3c/browser-specs/pull/504#issuecomment-1034024625

PR #528, opened 2022-02-24
Add new composition level: "fork"

tidoust:
This new composition level is meant for specs that are created as temporary forks of base specifications. First (and only so far) spec in that category is the "WebAssembly JavaScript Interface: Exception Handling" specification, as described in https://github.com/w3c/browser-specs/issues/508.

This will require a major version bump as this changes the semantics of the `seriesComposition` property.

(I'll work on an update to Reffy to handle fork specs)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,6 +55,9 @@ properties:
 - `url`: same as the [`url`](README.md#url) property in `index.json`.
 - `shortname`: same as the [`shortname`](README.md#shortname) property in
 `index.json`.
+- `forkOf`: same as the [`forkOf`](README.md#forkof) property in `index.json`.
+No need to set `seriesComposition` to `"fork"` when this property is set, the
+build logic will take care of that automatically.
 - `series`: same as the [`series`](README.md#series) property in `index.json`,
 but note the `currentSpecification` property will be ignored.
 - `seriesVersion`: same as the [`seriesVersion`](README.md#seriesversion)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,7 +54,9 @@ properties:
 
 - `url`: same as the [`url`](README.md#url) property in `index.json`.
 - `shortname`: same as the [`shortname`](README.md#shortname) property in
-`index.json`.
+`index.json`. When the `forkOf` property is also set, note that the actual
+shortname in the final list will be prefixed by the shortname of the base
+spec (as in: `${forkOf}-fork-${shortname}`).
 - `forkOf`: same as the [`forkOf`](README.md#forkof) property in `index.json`.
 No need to set `seriesComposition` to `"fork"` when this property is set, the
 build logic will take care of that automatically.
@@ -63,8 +65,9 @@ but note the `currentSpecification` property will be ignored.
 - `seriesVersion`: same as the [`seriesVersion`](README.md#seriesversion)
 property in `index.json`.
 - `seriesComposition`: same as the [`seriesComposition`](README.md#seriesComposition)
-property in `index.json`. The property must only be set for delta spec (since
-full is the default).
+property in `index.json`. The property must only be set for delta spec, since
+full is the default and fork specs are identified through the `forkOf` property
+in `specs.json`.
 - `organization`: same as the [`organization`](README.md#organization) property
 in `index.json` to specify the name of the organization that owns the spec.
 - `groups`: same as the [`groups`](README.md#groups) property in `index.json`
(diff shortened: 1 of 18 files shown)


specs.json
```diff
@@ -181,6 +181,11 @@
   "https://w3c.github.io/webappsec-trusted-types/dist/spec/",
   "https://w3c.github.io/webdriver-bidi/",
   "https://w3c.github.io/webrtc-ice/",
+  {
+    "url": "https://webassembly.github.io/exception-handling/js-api/",
+    "shortname": "wasm-js-api-1",
```
dontcallmedom, line 186: is it expected we would use the same shortname as a current spec? I was imagining we would mint a special one, while associating it with the series shortname (wasm-js-api).

src/lint.js
```diff
@@ -74,7 +79,8 @@ function lintStr(specsStr) {
       const next = linked.seriesNext ?
         linkedList.find(p => p.shortname === linked.seriesNext) :
         null;
-      const isLast = !next || next.seriesComposition === "delta";
+      const isLast = !next || next.seriesComposition === "delta" ||
```
dontcallmedom, line 82: probably not a problem today, but with the `fork` model, there isn't really a sense to "last" anymore, since there can be different parallel forks.

specs.json
tidoust: Right, that needs fixing. Tests won't pass once `index.json` gets built because two specs cannot have the same shortname. We'll need to mint a name à la `[base shortname]-fork-[fork-id]` and we may have to relax consistency tests which may yell that the spec's shortname and the series shortname seem mis-aligned.

src/lint.js
tidoust: That's true and that creates a problem. Today, we have `seriesPrevious` and `seriesNext` properties that create a linear set of specs in a series. If there is more than one fork of a spec, I don't know what `seriesNext` of the last spec that is not a fork in a series should point to. Nothing? Should there be a `forks` property to easily tell that a spec has forks and what shortname they have?

Sharon077 (approved):

src/lint.js
dontcallmedom: I think a short term solution would be not to mark forks as `seriesNext` (so in our case, wasm-js-api would always be the latest) and indeed document forks in a new property.

tidoust:
I updated the pull request to link fork and base specs through `forkOf` and `forks` properties. CSS server is currently down so I couldn't test an entire rebuild of the index but what could possibly go wrong? 😼

CONTRIBUTING.md
```diff
@@ -55,6 +55,9 @@ properties:
 - `url`: same as the [`url`](README.md#url) property in `index.json`.
 - `shortname`: same as the [`shortname`](README.md#shortname) property in
 `index.json`.
+- `forkOf`: same as the [`forkOf`](README.md#forkof) property in `index.json`.
+No need to set `seriesComposition` to `"fork"` when this property is set, the
+build logic will take care of that automatically.
```
dontcallmedom, line 60: maybe that should be mentioned explicitly in the `seriesComposition` entry (more specifically, IIUC, it is forbidden to set `fork` explicitly)

src/compute-shortname.js
```diff
@@ -142,42 +142,45 @@ function completeWithSeriesAndLevel(shortname, url) {
     }
   }
 
+  const seriesBasename = forkOf ?? shortname;
+  const specShortname = forkOf ? `${forkOf}-fork-${shortname}` : shortname;
```
dontcallmedom, line 146: it took me a while to walk the path that led to the conclusion that for the case at hand, `shortname` would have been calculated as `exception-handling` (which is great); I'm not sure where, but a comment that would make that easier to grasp would be helpful.

dontcallmedom (approved):
a couple of minor suggestions, but otherwise looks great!

src/compute-shortname.js
dontcallmedom: (all the more so that the fact that it worked for webassembly that uses separate repos for its forks may not remain universally true for this mode of composition)

CONTRIBUTING.md
tidoust: I completed the entry. It wasn't forbidden to set `fork` explicitly, actually. Now it is ;)

src/compute-shortname.js
tidoust: I completed the README and CONTRIBUTING files to explain the prefixing logic.

Sharon077 (approved):

PR #530, opened 2022-02-26
Add miniapp specs

xfq:
This PR enables "exported" terms in MiniApp specs to be automatically cross-referenced using https://respec.org/xref/

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -673,6 +673,24 @@
       "sourcePath": "MediaRecorder.bs"
     }
   },
+  {
+    "url": "https://www.w3.org/TR/miniapp-lifecycle/",
+    "categories": [
+      "-browser"
+    ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


xfq:
@dontcallmedom I just found that MiniApps Working Group was in the [ignore list](https://github.com/w3c/browser-specs/blob/c4bbab76ec65d2c80c40656d26299fabe1e6a8b1/src/data/ignore.json#L36), but it should be fine?

My guess is that it means that the specs do not (currently) target browsers, not that it is not ingested by this tool?

dontcallmedom:
> My guess is that it means that the specs do not (currently) target browsers, not that it is not ingested by this tool?

Indeed; more specifically, it means that the tool will not notify us when a new spec from the group emerges; that part of the infrastructure still needs to evolve in light of the broader scope of the project. It would probably makes sense to simply remove the group from that ignore list now that we've added its specs :)

PR #561, opened 2022-03-22
Throttle network requests

tidoust:
This adds throttling to network requests that used to be issued in parallel and that did not have explicit throttling, so as to be more gentle with web servers that the code needs to gather info from.

The `throttle` module exports a generic function that can be used to throttle another function to a maximum number of concurrent tasks. This module is used to throttle parallel requests to the W3C API server (in the hope that this can prevent 503 failures that we keep receiving), and to JSDOM (in the hope that being more gentle with the draft CSS server may help with its stability).

Code that uses Octokit now always goes through the throttled version as well.

Note that throttling is not used in `fetch-groups` simply because that module does not send requests in parallel for now (it uses a `for ... of ...` loop). We may actually want to speed up the module a bit later on through parallelism, but let's do that once build becomes more stable!

Creating a draft PR for now as I could not test the code thoroughly with an actual full re-build (build currently fails because the draft CSS server is down).

### DIFF
--- a/src/compute-repository.js
+++ b/src/compute-repository.js
@@ -7,7 +7,7 @@
  * The function needs an authentication token for the GitHub API.
  */
 
-const { Octokit } = require("@octokit/rest");
+const Octokit = require("./octokit");
 const parseSpecUrl = require("./parse-spec-url.js");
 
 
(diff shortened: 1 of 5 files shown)


tidoust:
I confirm that the code seems to work as intended. Note the parallelism is restricted to 2 for the time being. That can perhaps be relaxed later on, but the initial goal is to check whether this approach improves the current situation at all.

dontcallmedom (approved):

PR #565, opened 2022-03-27
Add WHATWG File System API

lukewarlow:
This PR adds https://fs.spec.whatwg.org/ to the spec list.

As part of this PR to MDN BCD: https://github.com/mdn/browser-compat-data/pull/15537


tidoust (approved):
Thanks! The corresponding entry also needs to be dropped from [`monitor.json`](https://github.com/w3c/browser-specs/blob/172f716e247ad0b744dde15eae1290d924cb99f4/src/data/monitor.json#L263). I'll do it in a separate pull request.

PR #605, opened 2022-05-10
WebRTC Extensions spec

hamishwillee:
This attempts to add the WebRTC extensions spec https://w3c.github.io/webrtc-extensions/ which is needed to allow me to update some of the properties in browser compatibility data: https://github.com/mdn/browser-compat-data/pull/16200

All I did was copy the previous record (WebRTC: Real-Time Communication Between Browsers) and attempt to update values. However I have not done this before and there were many values that I wasn't clear on. Advice on how to make this right welcome.

### DIFF
--- a/index.json
+++ b/index.json
@@ -17728,6 +17728,48 @@
     }
   },
   {
+    "url": "https://www.w3.org/TR/webrtc-extensions/",
+    "seriesComposition": "full",
+    "shortname": "webrtc",
+    "series": {
+      "shortname": "webrtc",
+      "currentSpecification": "webrtc",
+      "title": "WebRTC Extensions",
+      "shortTitle": "WebRTC Extensions",
+      "releaseUrl": "https://www.w3.org/TR/webrtc-extensions/",
+      "nightlyUrl": "https://w3c.github.io/webrtc-extensions/"
+    },
+    "shortTitle": "WebRTC Extensions",
+    "organization": "W3C",
+    "groups": [
+      {
+        "name": "Web Real-Time Communications Working Group",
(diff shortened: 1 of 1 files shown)


tidoust:
Thanks for the pull request, @hamishwillee.

We [did not include `webrtc-extensions`](https://github.com/w3c/browser-specs/blob/e1f233558108dae2d8d5db3fe0d7ad61bb63abd6/src/data/monitor.json#L39) so far because it is a kitchen-sink of IDL definitions that the WebRTC working group deems as not sufficiently mature to be part of the main spec. I guess we can revisit that decision. Ping @dontcallmedom who has a crisper vision of WebRTC specs.

FYI, pull requests need to touch `specs.json`, not `index.json`. This is described in the [contributing](https://github.com/w3c/browser-specs/blob/main/CONTRIBUTING.md#the-actual-list-is-in-specsjson) notes. I'm happy to redo the PR if we add the spec. The entry in the `monitor.json` file also needs to be dropped.

hamishwillee:
Ah. Thank you very much @tidoust . I have put an exception in the browser specs for now. I will try get a response from @dontcallmedom . At least part of this is now in both Safari and Firefox, so hopefully this means it is settling down a bit.

hamishwillee:
@tidoust Just an update from https://github.com/w3c/webrtc-extensions/issues/108 - you're right - that is a mixed bag of "not ready for the big time" methods and properties.
The plan is that approved bits get moved across to the proper spec. That is the plan for the particular property I am interested in.

So this change would not be made here. I'll close now and track in https://github.com/w3c/webrtc-extensions/issues/108

PR #647, opened 2022-07-06
Import Maps is now part of the HTML spec

dontcallmedom:
see https://github.com/whatwg/html/pull/8075


dontcallmedom:
making this is a draft pull request since the migration PR hasn't been merged yet

I filed this in advance because I'm not sure we would have a good way to detect the migration with our scripts (import maps doesn't expose new IDL, so we wouldn't detect e.g. duplicate interfaces)

tidoust (approved):

PR #667, opened 2022-07-28
Add/Monitor/Ignore new specs

### LINKED ISSUE #664: New specs for review

- [ ] https://tc39.es/proposal-duplicate-named-capturing-groups/ from [tc39/proposal-duplicate-named-capturing-groups](https://github.com/tc39/proposal-duplicate-named-capturing-groups)
- [ ] https://tc39.es/proposal-json-parse-with-source/ from [tc39/proposal-json-parse-with-source](https://github.com/tc39/proposal-json-parse-with-source)
- [ ] https://w3c.github.io/device-memory/ from [w3c/device-memory](https://github.com/w3c/device-memory)

tidoust:
This fixes #664

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -169,6 +169,7 @@
     }
   },
   "https://tc39.es/proposal-json-modules/",
+  "https://tc39.es/proposal-json-parse-with-source/",
   "https://tc39.es/proposal-resizablearraybuffer/",
   "https://tc39.es/proposal-shadowrealm/",
   {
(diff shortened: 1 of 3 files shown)


queengooborg:
Ignoring the Device Memory ED URL is causing problems over on our project (see https://github.com/mdn/browser-compat-data/pull/17148), as the TR URL does not seem to be included in this package either.

PR #681, opened 2022-08-22
Ignore older version of Payment Request

### LINKED ISSUE #680: New specs for review

- [ ] https://www.w3.org/TR/payment-request/ from [w3c/payment-request](https://github.com/w3c/payment-request) [chrome status](https://www.chromestatus.com/features/5948593429020672)
- [ ] https://www.w3.org/TR/payment-request/ from [w3c/payment-request](https://github.com/w3c/payment-request) [chrome status](https://www.chromestatus.com/features/5948593429020672)

tidoust:
Closes #680. False positive as list now contains v1.1 of the spec.

### DIFF
--- a/src/data/ignore.json
+++ b/src/data/ignore.json
@@ -381,9 +381,6 @@
     },
     "https://w3c.github.io/device-memory/": {
       "comment": "Already in the list. Code gets confused because TR spec links to itself as ED"
-    },
-    "https://www.w3.org/TR/payment-request/": {
-      "comment": "Older version of Payment Request 1.1"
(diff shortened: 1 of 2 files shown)


dontcallmedom:
hmm… why is this needed? Shouldn't the series & versioning take care of this automatically?

tidoust:
> hmm… why is this needed? Shouldn't the series & versioning take care of this automatically?

Someone may have written the find-specs script a bit hastily ;)

dontcallmedom (approved):

PR #684, opened 2022-08-25
Update all WebGL URLs

sideshowbarker:
All https://www.khronos.org/registry/ URLs now redirect to https://registry.khronos.org/ equivalents.


dontcallmedom:
good catch, thanks!

PR #688, opened 2022-08-30
Update HTTP Core spec URLs (RFCs 9110, 9111, 9112, 9113, 9114)

sideshowbarker:
I first tried just adding only the https://www.rfc-editor.org/rfc URLs for these — without also explicitly specifying the “nightly” URLs — but after running “npm run build”, I found that the build script doesn’t automatically pick up the https://httpwg.org/specs URLs as the nightly URLs.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -432,9 +432,7 @@
   {
     "url": "https://www.rfc-editor.org/rfc/rfc9112",
     "nightly": {
-      "url": "https://httpwg.org/specs/rfc9112.html",
-      "repository": "https://github.com/httpwg/httpwg.github.io",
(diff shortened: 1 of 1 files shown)


tidoust (changes requested):
Thanks for the PR and for digging into it :)

Once the nightly URL is set to an `httpwg.org` URL, browser-specs can automatically figure out the repository (hardcoded in [`parse-spec-url.js`](https://github.com/w3c/browser-specs/blob/ded5fefa28346ff3b36ee34307f12f0681cc3799/src/parse-spec-url.js#L63)), and from the repository it can find the source path as well. So no need to include them.

tidoust (approved):

PR #707, opened 2022-09-14
Add CSS Linked Parameters proposal

### LINKED ISSUE #705: New specs for review

- [ ] https://drafts.csswg.org/css-link-params/ from [w3c/csswg-drafts](https://github.com/w3c/csswg-drafts)

tidoust:
This used to be the "SVG Parameters" proposal. It got adopted by the CSS WG early August, with implementation interest from Mozilla: https://github.com/w3c/csswg-drafts/issues/6988#issuecomment-1203033606

Fixes #705

Creating as draft pull request as I cannot test the addition for now (CSS draft server is down)

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -40,6 +40,7 @@
     "shortTitle": "CSS GCPM 4"
   },
   "https://drafts.csswg.org/css-grid-3/ delta",
+  "https://drafts.csswg.org/css-link-params/",
   {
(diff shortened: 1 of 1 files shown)


tidoust:
Closing without merging as spec still hasn't been properly published on the CSS server. #719 adds the spec to the monitor list instead.

PR #711, opened 2022-09-20
Document change of nighlty shortname for CSS Viewport (née Device Adapt)

dontcallmedom:
See also https://github.com/w3c/csswg-drafts/issues/4766#issuecomment-585314599

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -557,7 +557,12 @@
   "https://www.w3.org/TR/css-contain-3/ delta",
   "https://www.w3.org/TR/css-content-3/",
   "https://www.w3.org/TR/css-counter-styles-3/",
-  "https://www.w3.org/TR/css-device-adapt-1/",
+  {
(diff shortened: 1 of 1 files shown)


specs.json
```diff
@@ -557,7 +557,12 @@
   "https://www.w3.org/TR/css-contain-3/ delta",
   "https://www.w3.org/TR/css-content-3/",
   "https://www.w3.org/TR/css-counter-styles-3/",
-  "https://www.w3.org/TR/css-device-adapt-1/",
+  {
+    "url": "https://www.w3.org/TR/css-device-adapt-1/",
+    "nightly": {
+      "url": "https://drafts.csswg.org/css-viewport/"
```
tidoust, line 563: The folder in the repo does not have a level, but the spec does, so suggest to be explicit about it.

```suggestion
      "url": "https://drafts.csswg.org/css-viewport-1/"
```

PR #712, opened 2022-09-20
Use uppercase HEAD in diff script

dontcallmedom:
branch names are case sensitive (in linux at least)


tidoust (approved):
Ah, interesting. Not in Windows as far as I can tell!

PR #726, opened 2022-09-27
Test unicity of sourcePath and fix duplication in SVG11 entry

### LINKED ISSUE #418: Add tests for duplicated `sourcePath`s
No spec should have same sourcePath (from #417).

tidoust:
This adds a test to ensure that we cannot have two different nightly versions of a spec that target the same repo and source path.

This typically happened with SVG11 and SVG2. The SVG11 entry incorrectly reported that its source is to be found in the w3c/svgwg repository. This update drops the association with that repository... and updates the underlying code in `compute-repository` not to take for granted that hypothetical repositories returned by `parseSpecUrl` actually exist.

This fixes #418.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -893,8 +893,7 @@
     "url": "https://www.w3.org/TR/SVG11/",
     "multipage": true,
     "nightly": {
-      "url": "https://www.w3.org/TR/SVG11/",
-      "repository": "https://github.com/w3c/svgwg"
+      "url": "https://www.w3.org/TR/SVG11/"
     }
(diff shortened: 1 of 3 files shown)


tidoust:
Grmpf, forgot that the change could make other tests fail. I'll look into it.

tidoust:
CI tests now fail because the new index file (without the sourcePath duplication) is not part of the PR. That's normal and proves the test work as intended :)

dontcallmedom (approved):

PR #738, opened 2022-10-10
Adjust URL of Resource Timing, ignore Image Maps

### LINKED ISSUE #736: New specs for review

- [ ] https://html.spec.whatwg.org/multipage/webappapis.html#import-maps from [WICG/import-maps](https://github.com/WICG/import-maps)

tidoust:
- Resource Timing no longer uses levels.
- Image Maps got folded into HTML and now redirects to it.

This fixes #736, #737.


dontcallmedom (approved):
maybe with s/image/import in the commit message

PR #763, opened 2022-11-02
Add RFC8288 support

sofalse:
[RFC8288](https://www.rfc-editor.org/rfc/rfc8288) is now a proposed standard, would be nice to have it here. Contents of the RFC (Link header) already implemented in Chrome.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -434,6 +434,7 @@
   "https://www.rfc-editor.org/rfc/rfc7725",
   "https://www.rfc-editor.org/rfc/rfc7838",
   "https://www.rfc-editor.org/rfc/rfc8246",
+  "https://www.rfc-editor.org/rfc/rfc8288",
   "https://www.rfc-editor.org/rfc/rfc8470",
   "https://www.rfc-editor.org/rfc/rfc8942",
   {
--- a/specs.json
+++ b/specs.json
@@ -434,7 +434,15 @@
   "https://www.rfc-editor.org/rfc/rfc7725",
   "https://www.rfc-editor.org/rfc/rfc7838",
   "https://www.rfc-editor.org/rfc/rfc8246",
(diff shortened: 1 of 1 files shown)


specs.json
```diff
@@ -434,6 +434,7 @@
   "https://www.rfc-editor.org/rfc/rfc7725",
   "https://www.rfc-editor.org/rfc/rfc7838",
   "https://www.rfc-editor.org/rfc/rfc8246",
+  "https://www.rfc-editor.org/rfc/rfc8288",
```
tidoust, line 437: This spec is interesting because it destroys one of our assumptions: that all specs would be developed under the umbrella of some group. Apparently, this one was developed by Mark Nottingham as an individual before it got adopted as a standard. The `source` field in https://www.rfc-editor.org/in-notes/rfc8288.json) is `"IETF - NON WORKING GROUP"`. Our code loses itself when it tries to compute the URL of that non-group.

On top of my head, three options:
1. Relax the schema not to require a non-empty `groups` property
2. Pretend that the group responsible for the spec is the IETF as a whole
3. Make Mark a group, flagging it as "individual in art area" as done in https://datatracker.ietf.org/doc/rfc8288/

I suggest the third option, leading to:

```suggestion
  {
    "url": "https://www.rfc-editor.org/rfc/rfc8288",
    "groups": [
      {
        "name": "Mark Nottingham (individual in art area)",
        "url": "https://datatracker.ietf.org/person/mnot@mnot.net"
      }
    ]
  },
```

tidoust (changes requested):

specs.json
sofalse: That third option also convinces me the most, seems like the logical approach to this problem. Updating the code to the proposed one.

tidoust (approved):

PR #775, opened 2022-11-17
Add https://www.w3.org/TR/test-methodology/

darobin:
I believe that this document is eligible under criterion 4, specifically "*it contains informative content that other specs often need to refer to (e.g. guidelines from horizontal activities such as accessibility, internationalization, privacy and security)*." I discovered that it was missing by trying to reference the `dfn` for `testable assertion` which I believe to be a rather central concept.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -934,6 +934,7 @@
     }
   },
   "https://www.w3.org/TR/SVG2/ multipage",
+  "https://www.w3.org/TR/test-methodology/",
   "https://www.w3.org/TR/timing-entrytypes-registry/",
   "https://www.w3.org/TR/touch-events/",
   {


dontcallmedom:
I think adding the doc to the web-specs list is uncontroversial; unfortunately, doing that in the current form of the Note won't get you what you want, since the Note doesn't use the `data-export` convention that is expected before a definition shows up in xref via webref. (I can't even blame the incompetence of the editors of that document since these conventions didn't exist back then)

That could be fixed in several ways:
* republishing the Note with improved markup; I *think* adding data-export could be construed as a [class 1 change](https://www.w3.org/2021/Process-20211102/#correction-classes) and so [could be done by the Team](https://www.w3.org/2021/Process-20211102/#publishing-notes), or if it's not, it could be done as a candidate correction.
* find a new group to own (and possibly republish) the document; I could imagine the Spec Editors Community Group being the right place to start an editors draft (although I'm not sure we have a clear process to authorize a CG to take over maintenance of a WG Note); just having something that can be documented as an editors draft would suffice in getting the right data extracted down the line

darobin:
Damn, foiled at every step!

If people _want_ to start working on this document again, I certainly would support that. However, apart from the examples and the format that are quite dated, does it need much change? I suspect it can stay as is, and so a class 1 change would seem to be the better avenue. I can provide an updated HTML source if it help?

One alternative I thought of: we could have a `legacy` mode for extraction, defined as just grabbing all the `dfn`. Then we could have `"https://www.w3.org/TR/test-methodology/ legacy"` in there and just make changes to the toolchain?

tidoust:
Some sort of `legacy` mode was also what came to mind when I read this. I created https://github.com/w3c/reffy/issues/1115 to track this in Reffy. On top of my head, I would rather suggest something slightly more specific than a `legacy` that does not really say much as to what we consider to be legacy, as in:

```json
{
 "url": "https://www.w3.org/TR/test-methodology/",
 "dfns": "tags only"
}
```

darobin:
That works for me just as well!

tidoust:
No need to flag the spec in any way in browser-specs after all. Next version of Reffy will [automatically detect legacy specs for dfns](https://github.com/w3c/reffy/pull/1120) and export dfns from legacy specs.

darobin:
This is excellent — thanks @tidoust! _Hannibal voice, chewing on a cigar_: I love it when a plan comes together without even having to change anything on my side!

PR #779, opened 2022-11-25
Add rdf-canon and RDF specs it references

### LINKED ISSUE #776: New specs for review

- [ ] https://w3c.github.io/rdf-canon/ from [w3c/rdf-canon](https://github.com/w3c/rdf-canon)

tidoust:
Also adds these specs (and test-methodology) to the list of specs for which we should not expect to find any repository or source path, because they don't have any.

This would fix #776.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -883,6 +883,15 @@
   "https://www.w3.org/TR/mixed-content/",
   "https://www.w3.org/TR/motion-1/",
   "https://www.w3.org/TR/mst-content-hint/",
+  {
+    "url": "https://www.w3.org/TR/n-quads/",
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):
if we're going to import more "legacy" specs moving forward, it may be worth making the "noRepo" rule part of the overridable data

tidoust:
Yes, I did not want to introduce too many changes here, but I'm thinking that we might actually want to stop testing for the presence of a `repository` and `sourcePath` properties altogether. At a minimum, the test on `sourcePath` does not make any sense when we already know that there is no `repository`. I created #781 to track this.

PR #808, opened 2023-01-05
Add new "status" property to "nightly" and "release"

tidoust:
This makes the code also report on the status of the nightly and release versions of each spec in the list.

This is intended to be used for analysis, but also to flag specs automatically as "unofficial" afterwards (planned for a separate PR).

Currently, that will create the following statuses:

Release statuses:
- Candidate Recommendation Draft: 33 specs
- Candidate Recommendation Snapshot: 29 specs
- Draft Note: 14 specs
- Draft Registry: 1 specs
- First Public Working Draft: 32 specs
- Note: 3 specs
- Proposed Recommendation: 1 specs
- Recommendation: 27 specs
- Retired: 2 specs
- Working Draft: 131 specs

Nightly statuses:
- A Collection of Interesting Ideas: 2 specs
- Draft Community Group Report: 63 specs
- Editor's Draft: 401 specs
- Experimental: 2 specs
- Informational: 1 specs
- Internet Standard: 3 specs
- Living Standard: 18 specs
- Proposed Standard: 26 specs
- TAG Finding: 1 specs
- Unofficial Proposal Draft: 1 specs

Interestingly, this shows that we have two retired specs in the list: "Do not track" and "Encoding" (the latter one, we should probably update to directly use the WHATWG URL instead of the /TR URL).

It also shows that we have two collections of interesting ideas and one unofficial draft already in the list: Web Animations 2 (which should really move to another status as noted in https://github.com/w3c/browser-specs/pull/290), CSS Typed OM 2 and Font Metrics API which we may have added to the list a bit prematurely (interfaces defined in the API are already in Web Platform Tests though).

Status of IETF specs is always "Editor's Draft" when the spec is not in Specref. That could perhaps be improved later on (and "Internet Standard" specs could perhaps be reported under `release` instead of under `nightly`.

There are few other statuses that we don't currently have in the list, notably those for more advanced registries. Tests will fail once they start appearing but that will provide a good occasion to manually check that all is fine.

The "TAG Finding" spec should actually rather be a "Draft Finding" but that's what Specref reports for now for that spec.

### DIFF
--- a/schema/definitions.json
+++ b/schema/definitions.json
@@ -67,13 +67,28 @@
       "type": "object",
       "properties": {
         "url": { "$ref": "#/$defs/url" },
+        "status": {
+          "type": "string",
+          "enum": [
+            "Candidate Recommendation Draft",
+            "Candidate Recommendation Snapshot",
+            "Draft Note",
+            "Draft Registry",
+            "First Public Working Draft",
+            "Note",
+            "Proposed Recommendation",
+            "Recommendation",
+            "Retired",
+            "Working Draft"
+          ]
+        },
         "filename": { "$ref": "#/$defs/filename" },
         "pages": {
           "type": "array",
           "items": { "$ref": "#/$defs/url" }
         }
       },
-      "required": ["url"],
+      "required": ["url", "status"],
       "additionalProperties": false
     },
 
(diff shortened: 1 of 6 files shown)


tidoust:
Forgot to add: status of the Patent Policy and of the Process Document will be "Editor's Draft", which does not seem fantastic... but I don't have any better status at hand a better status at hand, so I suggest to lazily ignore that for now.

schema/definitions.json
```diff
@@ -67,20 +67,51 @@
...
+          "enum": [
+            "Candidate Recommendation Draft",
+            "Candidate Recommendation Snapshot",
+            "Draft Note",
+            "Draft Registry",
+            "First Public Working Draft",
+            "Note",
+            "Proposed Recommendation",
+            "Recommendation",
+            "Retired",
```
dontcallmedom, line 81: Maybe we should align "Retired" with the W3C Process term, [Discontinued Draft](https://www.w3.org/2021/Process-20211102/#discontinuedREC)

tidoust: I had kept "Retired", and am tempted to keep it, because that is what we use on /TR pages and the status that the W3C API returns:
https://www.w3.org/standards/history/encoding

dontcallmedom: right - but my recollection is that I coined that term "Retired" out of thin air some 20 years ago in the context of modeling TR in RDF, and that trickled down into the 2008 Web site redesign and later on in the W3C API. Now that we have a better consensus-based name, I don't think there is actual value in sticking to that accident of history :)

dontcallmedom (approved):

PR #812, opened 2023-01-06
Add "unofficial" category

tidoust:
This update creates a new `unofficial` category for specs whose status is "A Collection of Interesting Ideas" or "Unofficial Proposal Draft".

As usual, default rule can be overridden in `specs.json`. Done for font-metrics-api-1 and web-animations-2 because they define IDL that has already been included in Web Platform Tests.

The css-typed-om-2 spec will be the only one to get an `unofficial` flag for now. It deserves it, the spec is mostly empty for now (we could just as well drop the spec, but since it's been in the list since the start, let's keep it there).

Unofficial specs appear in the web-specs package but get filtered out of the browser-specs package.

The new status will make it possible to add a few additional unofficial CSS specs. To be done in a separate commit.

### DIFF
--- a/schema/definitions.json
+++ b/schema/definitions.json
@@ -167,22 +167,21 @@
       "type": "array",
       "items": {
         "type": "string",
-        "enum": ["browser"]
+        "enum": ["browser", "unofficial"]
       }
+    },
+
+    "categories-delta": {
+      "type": "string",
+      "enum": ["reset", "+browser", "-browser", "+unofficial", "-unofficial"]
     },
 
     "categories-specs": {
       "oneOf": [
-        {
-          "type": "string",
-          "enum": ["reset", "+browser", "-browser"]
-        },
+        { "$ref": "#/$defs/categories-delta" },
         {
           "type": "array",
-          "items": {
-            "type": "string",
-            "enum": ["reset", "+browser", "-browser"]
-          },
+          "items": { "$ref": "#/$defs/categories-delta" },
           "minItems": 1
         }
       ]
--- a/specs.json
+++ b/specs.json
@@ -56,7 +56,12 @@
   },
   "https://dom.spec.whatwg.org/",
   "https://drafts.css-houdini.org/css-typed-om-2/ delta",
-  "https://drafts.css-houdini.org/font-metrics-api-1/",
+  {
+    "url": "https://drafts.css-houdini.org/font-metrics-api-1/",
+    "categories": [
+      "-unofficial"
+    ]
+  },
   "https://drafts.csswg.org/css-anchor-1/",
   "https://drafts.csswg.org/css-animations-2/ delta",
   {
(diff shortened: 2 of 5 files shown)


schema/definitions.json
```diff
@@ -167,22 +167,21 @@
       "type": "array",
       "items": {
         "type": "string",
-        "enum": ["browser"]
+        "enum": ["browser", "unofficial"]
```
dontcallmedom, line 170: I don't think we should re-use categories for this - this is more of a status thing than a category; if we were to start other categories (e.g. linked data, verifiable credentials, mini-apps), they could also have unofficial drafts that we may want to handle differently.

tidoust:
Note that a "negative" flag is usually not a fantastic idea, but I'm somewhat expecting unofficial entries to remain an exception to the rule, so I stuck to "unofficial". Other names could be considered. For instance, "proposal" (but then many people might consider WICG specs to be "proposals"). Ideas welcome ;)

schema/definitions.json
tidoust: My problem is that we'll want to override what we want to consider "unofficial" in some cases, and I don't think we should override the status of the spec. How would you like to do that?

dontcallmedom: I think we want a separate entry that can be used to filter whether / where the extracted data is taken into account; if we focus on status, it could be something labeled e.g. "standing" (?); or maybe we're really talking about implementation level, in which case the label could reflect that?

tidoust: What we want to achieve strikes me as ill-defined. I see how we could call specs we'd want to exclude from browser-specs ("unofficial", "idea", "proposal") but I don't know how to call specs we want to keep ("official" feels too strong, "implementers expressed interest" somewhat arbitrary). Hence the use of "categories" which, I agree, is not ideal, but allows to use only one term and remain implicit on what it means when this term is missing ;)

We could perhaps override the "status" property after all. That is, we could say that we consider font-metrics-api-1 and web-animations-2 to be proper drafts. One example of spec for which I find this approach somewhat clunky would be [Web Monetization](https://webmonetization.org/specification.html), which is currently on our monitoring list. Status is CG draft. The spec is explicit that it is "work in progress" but, if we add it to the list, calling it "Unofficial Proposal Draft" feels a tad too strong.

Or we create a `standing` property whose meaning remains a bit fuzzy and whose value can either be:
- `good`, meaning spec has some level of standing.
- `pending` or `ongoing`, meaning spec should still be regarded as unofficial
How does that look?

dontcallmedom: I'm fine with your `standing` proposal; the more I think about it, the more this seems to be about implementation interest, but my guess is that we can start with `standing` and make it evolve when things get less fuzzy as we start using it.

tidoust: OK, done in #813!

PR #813, opened 2023-01-06
Add new "standing" property

tidoust:
Second take at #812.

This update creates a new `standing` category for specs whose status is "A Collection of Interesting Ideas" or "Unofficial Proposal Draft".

As usual, default rule can be overridden in `specs.json`. Done for font-metrics-api-1 and web-animations-2 because they define IDL that has already been included in Web Platform Tests.

The css-typed-om-2 spec will be the only one to get a `"pending"` standing for now.

Specs in pending status appear in the web-specs package but get filtered out of the browser-specs package.

The new status will make it possible to add a few additional unofficial CSS specs. To be done in a separate commit.

### DIFF
--- a/src/compute-standing.js
+++ b/src/compute-standing.js
@@ -1,0 +1,34 @@
+/**
+ * Module that exports a function that takes a spec object that already has its
+ * `nightly.status` (and `release.status` for released specs) properties set as
+ * input, and that returns the "standing" of the spec.
+ *
+ * Note (2023-01-06): The definition of "standing" remains fuzzy and this
+ * property should be regarded as unstable.
+ */
+
+// List of spec statuses that are not "official" ones, in the sense that the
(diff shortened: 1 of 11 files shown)


src/compute-standing.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Exports main function that takes a spec object and returns the standing of
+ * the spec.
+ */
+module.exports = function (spec) {
+  if (!spec || !spec.nightly?.status) {
+    throw "Invalid spec object passed as parameter";
+  }
+
+  // If spec is already explicit about its standing, who are to disagree?
```
dontcallmedom, line 27:
```suggestion
  // If spec is already explicit about its standing, who are we to disagree?
```

dontcallmedom (approved):
with a nit

PR #824, opened 2023-01-09
Ignore current spec from W3C API if series is different

tidoust:
The currently list asserts that `css-2022` is the current specification of the `CSS` series, even though, for us, the `CSS` series is different from the `css` series (the former contains the CSS 2.x specs, the latter contains the CSS snapshots).

This is because the W3C API considers that these series are just one series and returns `css-2022` as the current specification. This update makes sure that the code ignores the current specification returned by the W3C API when it does not belong to the same specification series for us.

This will make the code get back to `CSS22` as the current specification for the `CSS` series.


tidoust:
Merging to fix data curation in Webref and the CSS extract of the CSS spec.

Issue #837, opened 2023-01-17
css-device-adapt is renamed to css-viewport


tabatkins:
The older CSS Device Adaptation spec was retired, and most of its content moved to a new CSS Viewport spec. The Reffy data correctly recognizes that <https://drafts.csswg.org/css-viewport/> is the ED for <https://www.w3.org/TR/css-device-adapt-1/>, but still has them categorized under the "css-device-adapt-1" shortname.

At some point in the near future (should happen this year) we'll publish a /TR of css-viewport, thus completely replacing the older device-adapt name. Can we change the shortname in the data now?

(This was reported by a Bikeshed user trying to link to a section in css-viewport and getting a "no spec of that name" error.)

tidoust:
"css-device-adapt-1" is now gone, replaced by "css-viewport". The update has already cascaded to bikeshed-data as far as I can tell.

PR #838, opened 2023-01-18
Replace css-device-adapt-1 with css-viewport

tidoust:
The css-device-adapt-1 spec will be retired soon and replaced by css-viewport. We had already updated the entry to have css-viewport be the Editor's Draft of css-device-adapt-1 but that remains confusing. This update completely drops css-device-adapt-1.

See https://github.com/w3c/browser-specs/issues/837

New entry would be:

```json
{
  "url": "https://drafts.csswg.org/css-viewport/",
  "seriesComposition": "full",
  "shortname": "css-viewport",
  "series": {
    "shortname": "css-viewport",
    "currentSpecification": "css-viewport",
    "title": "CSS Viewport",
    "shortTitle": "CSS Viewport",
    "nightlyUrl": "https://drafts.csswg.org/css-viewport/"
  },
  "organization": "W3C",
  "groups": [
    {
      "name": "Cascading Style Sheets (CSS) Working Group",
      "url": "https://www.w3.org/Style/CSS/"
    }
  ],
  "nightly": {
    "url": "https://drafts.csswg.org/css-viewport/",
    "status": "Editor's Draft",
    "alternateUrls": [
      "https://w3c.github.io/csswg-drafts/css-viewport/"
    ],
    "repository": "https://github.com/w3c/csswg-drafts",
    "sourcePath": "css-viewport/Overview.bs",
    "filename": "Overview.html"
  },
  "title": "CSS Viewport Module Level 1",
  "source": "spec",
  "shortTitle": "CSS Viewport 1",
  "categories": [
    "browser"
  ],
  "standing": "good"
}
```

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -120,6 +120,7 @@
     "url": "https://drafts.csswg.org/css-variables-2/",
     "shortTitle": "CSS Variables 2"
   },
+  "https://drafts.csswg.org/css-viewport/",
   {
     "url": "https://drafts.csswg.org/web-animations-2/",
     "seriesComposition": "delta",
@@ -663,13 +664,6 @@
   "https://www.w3.org/TR/css-contain-3/ delta",
   "https://www.w3.org/TR/css-content-3/",
   "https://www.w3.org/TR/css-counter-styles-3/",
-  {
-    "url": "https://www.w3.org/TR/css-device-adapt-1/",
-    "nightly": {
-      "url": "https://drafts.csswg.org/css-viewport-1/",
-      "sourcePath": "css-viewport/Overview.bs"
-    }
-  },
   "https://www.w3.org/TR/css-display-3/",
   "https://www.w3.org/TR/css-easing-1/",
   {
--- a/src/data/ignore.json
+++ b/src/data/ignore.json
@@ -384,9 +384,6 @@
     },
     "https://www.w3.org/TR/resource-timing-1/": {
       "comment": "Level-less spec is in the list, no need to track old Level 1 spec"
-    },
-    "https://www.w3.org/TR/css-device-adapt-1/": {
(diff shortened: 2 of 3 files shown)


dontcallmedom:
couldn't we override the shortname instead? I assume TR/css-viewport will be published as a successor to css-device-adapt, and I'm not sure how this will impact detection with css-device-adapt in ignore.json?

tidoust:
What does "successor" mean here and why keep the /TR spec?

I assumed TR/css-device-adapt would get published as a Discontinued Draft, and TR/css-viewport would be created as a new draft. Is there a /TR mechanism in place to represent the notion of successor?

dontcallmedom:
There is a mechansim to change a TR shortname; it would be a successor in the sense that TR/YYYY/WD-css-viewport-YYYYMMDD would have [TR/WD-css-device-adapt](http://www.w3.org/TR/2016/WD-css-device-adapt-1-20160329/) as previous version, and TR/css-device-adapt would redirect to TR/css-viewport

I'm not sure that this is how this will be handled, and maybe it wouldn't impact detection if it is, but `ignore.json` feels a bit blunt of a tool.

tidoust:
In the end, we'll want to end up with the entry URL `https://www.w3.org/TR/css-viewport/`, and no entry related to css-device-adapt. The entry in `ignore.json` will become useless, and whatever we do today, we'll have to revisit what's in `specs.jsn`.

Are you worried that the entry in `ignore.json` will stay forever whereas we should drop it once the transition is over? If so, I can move the entry to `monitor.json`. Or are you worried that switching to css-viewport now is premature? If so, I can override the shortname, I just don't see the need to capture the /TR draft that already seems to be considered retired.

dontcallmedom:
yeah, my worry is mostly that we miss the transition to TR/css-viewport once it's published; using monitor.json SGTM

tidoust:
Deal! ;)

dontcallmedom (approved):

PR #848, opened 2023-01-24
currentSpecification value

estelle:
Updated  `currentSpecification` value to `compositing-2`.

I am not 100% sure my edit corrects the issues, but there is an issue that needs to be corrected.

There are 2 compositing specifications. 
  - https://drafts.fxtf.org/compositing/ (also at https://drafts.fxtf.org/compositing-2/)
  - https://www.w3.org/TR/compositing-1/ (also at https://drafts.fxtf.org/compositing-1/)

But both render the same (as "Compositing and Blending Level 1") when the JSON is parsed with tools such as Yari on MDN.

### DIFF
--- a/index.json
+++ b/index.json
@@ -1410,7 +1410,7 @@
     "shortname": "compositing-2",
     "series": {
       "shortname": "compositing",
-      "currentSpecification": "compositing-1",
+      "currentSpecification": "compositing-2",
       "title": "Compositing and Blending",
       "shortTitle": "Compositing",
       "releaseUrl": "https://www.w3.org/TR/compositing/",


tidoust:
Thanks for reporting the issue, @estelle. If I understand it correctly, browser-specs data is inconsistent because it claims that `currentSpecification` is level 1 but points at https://drafts.fxtf.org/compositing/ as the nightly URL for the series, which actually returns level 2.

This problem exists because the release URL for the series is https://www.w3.org/TR/compositing/, which currently returns level 1 (Level 2 has not yet been published to /TR). The CSS Working Group usually aligns the two.

Forcing `currentSpecification` to Level 2 would create the opposite inconsistency: the release URL for the series would return level 1 instead of level 2. I'll try to come up with a more proper fix (assuming that the CSS Working Group indeed intends to have Level 2 as the current spec).

FYI, the `index.json` is generated from `specs.json` file, so pull requests should not touch `index.json`.

tidoust:
Done in #851 and new versions of NPM packages just released with the update.

> Forcing currentSpecification to Level 2 would create the opposite inconsistency: the release URL for the series would return level 1 instead of level 2.

That consistency is not really a problem upon further reflection. The release URL returns level 1 but then that is the latest level being published, so that's correct!

estelle:
Thank you!

PR #854, opened 2023-01-27
Add requestStorageAccessForOrigin and First-Party Sets specs

johannhof:
Let me know if I missed anything here :)


tidoust:
Thanks, @johannhof!

First-Party Sets is in our [monitor list](https://github.com/w3c/browser-specs/blob/35ae5f572d5e2348508408bd2d49fd23e0a50b65/src/data/monitor.json#LL23-L26) and we noted negative signals from Mozilla and Apple earlier this month. Unless something changed, I'm wondering whether to add it with a [`standing`](https://github.com/w3c/browser-specs/#standing) property set to `pending` for now. @dontcallmedom, WDYT?

Additional comments:
- If we add First-party Sets to the list, the corresponding entry needs to be dropped from `monitor.json`.
- Code needs the `w3c.json` file to extract the group's name. Build won't work until it gets added: https://github.com/privacycg/requestStorageAccessForOrigin/pull/20

johannhof:
Thanks for submitting that PR!

To your question, I'm not 100% sure what qualifies as "pending" but Chrome has implemented FPS and will most likely ship it, so it's definitely more than "a collection of interesting ideas" from our POV. If you have a precedent for how other contentious APIs with strong commitment from a single implementer were handled in the past, you should apply it here :)

I'll go ahead and drop the entry from `monitor.json` then.

dontcallmedom:
@johannhof when looking up [FPS on Chrome status](https://chromestatus.com/feature/5640066519007232), it points to https://bugs.chromium.org/p/chromium/issues/detail?id=1117249 which shows very little activity. Can you point to more information on implementation & shipping plans?

johannhof:
Hi @dontcallmedom, yeah, that bug is a little stale, sorry. There's more activity visible on https://bugs.chromium.org/p/chromium/issues/list?q=component:Internals%3ENetwork%3EFirst-Party-Sets in terms of implementation. I can't commit to a specific milestone for shipping but we've been getting things into shape to reach the level of maturity that is required by the Blink launch process (including the spec linked here and TAG/browser positions), so I hope that we'll be able to send an intent to ship within the next few releases.

dontcallmedom:
thanks @johannhof! I think that's a good match for the [criteria](https://github.com/w3c/browser-specs/tree/main/packages/browser-specs#spec-selection-criteria) of 
> Web browsers expressed some level of support for the spec, e.g. through a public intent to implement.

I would thus lean towards merging it with a default standing (i.e. as-is in this PR)

tidoust (approved):
Then we're all set! Build passes. Merging now.

johannhof:
Thanks both!

PR #897, opened 2023-03-06
Force nightly status of CSS Animations 2 to ED

tidoust:
The status of the nightly version of CSS Animations 2 is currently set to "First Public Working Draft". Not such a big deal, but not one of the expected values per our JSON schema.

If this type of issues keeps re-appearing, we may want to do one of:
1. add logic to convert the status of nightly versions to "Editor's Draft" when needed
2. relax the JSON schema to also accept release statuses
3. keep on being strict and report issues to groups

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -76,13 +76,7 @@
     "standing": "good"
   },
   "https://drafts.csswg.org/css-anchor-position-1/",
-  {
-    "url": "https://drafts.csswg.org/css-animations-2/",
-    "seriesComposition": "delta",
-    "nightly": {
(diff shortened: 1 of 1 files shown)


dontcallmedom (changes requested):
per #896, css-animations-2 has reached [FPWD](https://www.w3.org/TR/css-animations-2/), so wouldn't the right fix rather to move to use the TR url?

tidoust:
That's because I'm doing things one step at a time. First, fix the build, then refresh ;)

dontcallmedom (approved):

PR #938, opened 2023-04-27
Adjust short title for Protected Audience

tidoust:
Title of the `turtledove` proposal changed to:

  "Protected Audience (formerly FLEDGE)"

The code thinks that the text between parentheses is the short title since that's the usual convention. This update hardcodes the short title to "Protected Audience".

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -573,7 +573,10 @@
   },
   "https://wicg.github.io/speech-api/",
   "https://wicg.github.io/trust-token-api/",
-  "https://wicg.github.io/turtledove/",
+  {
+    "url": "https://wicg.github.io/turtledove/",
+    "shortTitle": "Protected Audience"
+  },
   "https://wicg.github.io/ua-client-hints/",
   "https://wicg.github.io/urlpattern/",
   "https://wicg.github.io/user-preference-media-features-headers/",


dontcallmedom (approved):

domfarolino:
How can you refer to this from specs? I'm trying `[[Protected Audience]]` as well as `[[Protected-Audience]]` and getting no luck linking to the specification.

tidoust:
The database that gets used to resolve biblio references is [Specref](https://www.specref.org/). We don't (yet?) have a mechanism in place to push specs added to browser-specs to Specref automatically and it does not seem to know anything about Protected Audience for now. You may request the spec to be [added to Specref](https://github.com/tobie/specref#updating--adding-new-references). Alternatively, you may add a [custom biblio entry](https://speced.github.io/bikeshed/#biblio-custom) to the spec that needs to reference Protected Audience.

domfarolino:
I see. I'll consider adding the spec to that repository then. May I ask what the difference between this one and that one is? What impact on specifications does the giant list of specs _in this repo_ have?

tidoust:
> May I ask what the difference between this one and that one is?

The scope of the list of browser-specs is essentially latest versions of specs that are part of the web platform and that can be crawled automatically to extract meaningful information (IDL, definitions, headings, etc.). Specref has a broader scope in that it contains all published versions of a given spec and lists specs that cannot be usefully crawled for now (e.g., because they don't follow specific semantic patterns to define terms, or because they were published many years ago when these patterns did not exist).

Info in browser-specs comes from Specref (or from the sources that Specref uses) when possible. Browser-specs contains info that Specref does not have such as links between versions of specs in the same series. For instance, Specref has all css-fonts specs, but does not tell you that css-fonts-3, css-fonts-4 and css-fonts-5 are related, that css-fonts-4 is the current one today and that css-fonts-5 is a delta specification.

We tend to add specs to browser-specs sooner than they get added in Specref because the need to parse and validate a spec arises sooner than the need to reference them.

Instead of using Specref as a source, browser-specs could actually become a source for Specref. That's what my "(yet?)" was meant to acknowledge ;)

> What impact on specifications does the giant list of specs in this repo have?

This repo sets the list of specifications that gets crawled by [Reffy](https://github.com/w3c/reffy) every 6 hours to populate data in https://github.com/w3c/webref, which includes a cross-reference database (the one used in Bikeshed and ReSpec), IDL extracts (used in Web Platform Tests and in TypeScript), and CSS grammar extracts (used in MDN).

tidoust:
> Instead of using Specref as a source, browser-specs could actually become a source for Specref.

Now done. Contributing a spec to browser-specs should now be enough to have it appear in Specref too.

Issue #945, opened 2023-05-05
Consider adding Screen Capture CG to monitored CGs


dontcallmedom:
see https://github.com/w3c/browser-specs/pull/941#pullrequestreview-1414787125

this would also suggest ensuring the repos there get a proposer w3c.json

eladalon1983:
Hey @dontcallmedom. How can I help this consideration process? :-)

dontcallmedom:
@eladalon1983 this issue is about monitoring the Screen Capture CG deliverables overall; to get a specific spec added (à la https://github.com/screen-share/mouse-events/), filing an issue (or even a PR!) for the said spec is more likely to get you what you want sooner :)

eladalon1983:
What's the SCCG's group number?

dontcallmedom:
if this is for the w3c.json file, you don't need the group number any more, you can use `cg/sccg` per https://w3c.github.io/w3c.json.html

eladalon1983:
Thanks!

eladalon1983:
P.S: Contacts - my GitHub handle, or...?

eladalon1983:
(FYI emoji reactions do not trigger emails, so easy to miss.)

eladalon1983:
(Thanks for answering. FYI emoji reactions do not trigger emails, so easy to miss.)

fred-wang:
@dontcallmedom @tidoust Can we close that issue now?

dontcallmedom:
I think so - while there may still be value in having a more systematic monitoring, the fact there is a well-paved path for the SCCG specs to land here is sufficient from my perspective

Issue #944, opened 2023-05-05
Draft status of ReSpec specs may be missed


tidoust:
When it needs to retrieve information from the spec itself, the code uses JSDOM without running scripts. For ReSpec draft specs (that do not get published through the spec-prod action), this means that the code does not have access to the generated spec and in particular won't find the spec's status.

This prevents the code from extracting the actual spec's status, resorting to "Editor's Draft". Most of the time, that is correct, but we may need to add unofficial drafts and CG drafts from time to time.

Examples of specs for which that is an issue:
- [Direct Sockets API](https://wicg.github.io/direct-sockets/)
- [Element Capture](https://screen-share.github.io/element-capture/)

I don't know whether we can use JSDOM to generate a ReSpec spec (we switched to Puppeteer in Reffy partly because that did not work well)

PR #994, opened 2023-07-03
Add Element Capture and Captured Mouse Events specifications

fred-wang:
See https://github.com/w3c/browser-specs/issues/945

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -237,6 +237,8 @@
       ]
     }
   },
+  "https://screen-share.github.io/captured-mouse-events/",
+  "https://screen-share.github.io/element-capture/",
   {
     "url": "https://sourcemaps.info/spec.html",
     "shortname": "sourcemap",
--- a/src/data/monitor.json
+++ b/src/data/monitor.json
@@ -365,10 +365,6 @@
     "https://www.w3.org/TR/css-device-adapt-1/": {
       "lastreviewed": "2023-05-01",
       "comment": "spec should be retired in 2023, see https://github.com/w3c/browser-specs/issues/837, replaced by css-viewport"
-    },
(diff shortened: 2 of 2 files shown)


fred-wang:
cc @dontcallmedom @eladalon1983

This is a tentative PR (untested). I understand index.json is generated automatically, by the fetch-info bot so no need to touch it.

tidoust:
Seems good but `w3c.json` files in related repos are invalid JSON for now, see https://github.com/screen-share/captured-mouse-events/pull/22 and https://github.com/screen-share/element-capture/pull/15

tidoust:
Also, both drafts claim that they are "Unofficial drafts", which we usually don't add to the list. As far as I can tell, these are CG drafts. Could the status be updated to reflect that? (In ReSpec, that should be `CG-DRAFT`)

eladalon1983:
I've fixed that; could you PTAL?

fred-wang:
@tidoust Can you please check again?

specs.json
```diff
@@ -237,6 +237,8 @@
       ]
     }
   },
+  "https://screen-share.github.io/captured-mouse-events/",
+  "https://screen-share.github.io/element-capture/",
```
tidoust, line 241: Need to be explicit about status for now because of #944:

```suggestion
  {
    "url": "https://screen-share.github.io/captured-mouse-events/",
    "nightly": {
      "status": "Draft Community Group Report"
    }
  },
  {
    "url": "https://screen-share.github.io/element-capture/",
    "nightly": {
      "status": "Draft Community Group Report"
    }
  },
```

fred-wang: Thanks!

PR #1000, opened 2023-07-07
Send ReSpec specs to spec-generator before looking for status

### LINKED ISSUE #944: Draft status of ReSpec specs may be missed

tidoust:
Fixes #944. Note generation of ReSpec specs with JSDOM is still not possible in practice, due to missing APIs in JSDOM required by ReSpec.

To make sure that the result would work for all WICG specs (in case browser-specs becomes the source for Specref for WICG specs), I made a built without retrieving info from Specref for WICG specs. Result looks good, and would actually fix a couple of outdated titles and incorrect spec statuses.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -237,18 +237,8 @@
       ]
     }
   },
-  {
-    "url": "https://screen-share.github.io/captured-mouse-events/",
-    "nightly": {
-      "status": "Draft Community Group Report"
-    }
-  },
-  {
-    "url": "https://screen-share.github.io/element-capture/",
-    "nightly": {
-      "status": "Draft Community Group Report"
-    }
-  },
+  "https://screen-share.github.io/captured-mouse-events/",
+  "https://screen-share.github.io/element-capture/",
   {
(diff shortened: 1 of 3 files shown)


dontcallmedom (approved):
I'm not sure how reliable the on-line spec generator is expected to be; but I'll guess we'll see based on usage

PR #1013, opened 2023-07-25
Add step-by-step build mode with intermediary files

tidoust:
Build of the index file sometimes fails at intermediary steps and often fails at the testing phase. In both cases, a new build needs to be run locally to investigate more precisely what fails. That's tedious and takes time.

This update creates a "step-by-step" mode whereby the build records the result of each step to an intermediary file in the `buildsteps` folder. These files are only ever going to be useful for debugging purpose, but that's already a good thing.

The build job now pushes intermediary files that got updated to the repository no matter what, even when the build fails. This allows to restart the build locally from the first step that failed.

The `10-filename.json` file in the `buildsteps` folder contains the final results of the build (and thus matches the `index.json` when tests pass). That file should be particularly useful to spot what needs fixing for tests to pass.

A possible alternative would have been to use separate branches but that seemed much more complicated to setup and use.


dontcallmedom:
I haven't reviewed the changes in details - while I understand and support the goals, my initial reaction is that storing intermediate build files on the main branch is going to be pretty noisy…

tidoust:
@dontcallmedom, following your suggestion, the job now stores intermediary files as artefacts (which should be kept on GitHub for 90 days).

Internally, these files are always created and stored in a `.buildsteps` local folder, listed in `.gitignore`. I decided not to create a separate build mode after all as that does not seem needed.

tidoust:
> I assume we need to merge before we get to check it works as we expect and we get what we want from artifacts?

i-dont-always-test-but-when-i-do-i-test-in-production-304471-1

dontcallmedom:
seems to have worked alright :) https://github.com/w3c/browser-specs/suites/15481668017/artifacts/884091580

PR #1023, opened 2023-07-31
Remove W3C API key

dontcallmedom:
no longer required (not even recommended)


test/fetch-groups-gh.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * Tests for the fetch-groups module that require a W3C API key
+ *
+ * These tests are separated from the tests that do not require a W3C API key
```
tidoust, line 4:
```suggestion
 * Tests for the fetch-groups module that require a GitHub token
 *
 * These tests are separated from the tests that do not require a GitHub token
```

... but in practice, do we still need to keep these tests separated? They had to because of the W3C API key secret. However, if I read the documentation correctly, the GitHub token should be set on pull requests, even on pull requests that originate from forked repositories:

> For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork
> https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

package.json
```diff
@@ -15,7 +15,7 @@
     "lint": "node src/lint.js",
     "lint-fix": "node src/lint.js --fix",
     "test": "mocha",
-    "test-pr": "mocha --exclude test/fetch-info-w3c.js --exclude test/fetch-groups-w3c.js"
+    "test-pr": "mocha --exclude test/fetch-groups-gh.js"
```
tidoust, line 18: See comment on `fetch-groups-gh.js`. My understanding is that `test-pr` could be dropped as the job that runs on pull requests should now be able to run `test` directly.

tidoust (changes requested):

tidoust (approved):

Issue #1008, opened 2023-07-13
Tons of new specs


BenjaminAster:
I was working on my own browser spec collection for myself, using [w3c/groups](https://github.com/w3c/groups) as a starting point and manually adding lots of stuff, partially by looking at the interfaces in the window object in Chrome's JavaScript console &ndash; until I realized the existence of this repo and webref. This has the disadvantage of having wasted a lot of time, but also the advantage of being able to diff my spec list with yours and finding the differences.

Here are specs that I believe should be added to this repo:

- Topics API: https://patcg-individual-drafts.github.io/topics/
- Private Aggregation API: https://patcg-individual-drafts.github.io/private-aggregation-api/
- WebXR Marker Tracking Module: https://immersive-web.github.io/marker-tracking/
- WebXR Meshing API Level 1: https://immersive-web.github.io/real-world-geometry/webxrmeshing-1.html
- WebXR Plane Detection Module: https://immersive-web.github.io/real-world-geometry/plane-detection.html
- Capture all screens: https://screen-share.github.io/capture-all-screens/
- Audio Session: https://w3c.github.io/audio-session/
- execCommand: https://w3c.github.io/editing/docs/execCommand/
- Mathematical Markup Language (MathML) Version 4.0: https://w3c.github.io/mathml/
- Media Capture and Streams Extensions: https://w3c.github.io/mediacapture-extensions/
- The Capture-Handle Actions Mechanism: https://w3c.github.io/mediacapture-handle/actions/
- Media Source Extensions Byte Stream Format Registry: https://w3c.github.io/mse-byte-stream-format-registry/
- ISO BMFF Byte Stream Format: https://w3c.github.io/mse-byte-stream-format-isobmff/
- MPEG-2 TS Byte Stream Format: https://w3c.github.io/mse-byte-stream-format-mp2t/
- MPEG Audio Byte Stream Format: https://w3c.github.io/mse-byte-stream-format-mpeg-audio/
- WebM Byte Stream Format: https://w3c.github.io/mse-byte-stream-format-webm/
- Confinement with Origin Web Labels: https://w3c.github.io/webappsec-cowl/
- Suborigins: https://w3c.github.io/webappsec-suborigins/
- User Interface Security and the Visibility API: https://w3c.github.io/webappsec-uisecurity/
- WebRTC Extensions: https://w3c.github.io/webrtc-extensions/
[...]

PR #1038, opened 2023-08-30
Add Audio Session and MSE registry/registrations

tidoust:
Via https://github.com/w3c/browser-specs/issues/1008

All specs are developed/maintained by the Media WG.


dontcallmedom (approved):
was https://github.com/w3c/audio-session/commit/73aab448a8821f77054c809ec68ebaf445552670 the reason why we didn't detect audio-session?

tidoust:
> was [w3c/audio-session@73aab44](https://github.com/w3c/audio-session/commit/73aab448a8821f77054c809ec68ebaf445552670) the reason why we didn't detect audio-session?

Yes. FWIW, I got a bit lost because the [validate-repo report](https://w3c.github.io/validate-repos/report.html) does not seem to list these group errors.

dontcallmedom:
I've now added some early semi-automatic reporting of validate-repos errors, directly inspired from what we have in strudy FWIW

PR #1043, opened 2023-09-01
Add more TC39 proposals at stage 3 + Source Maps

tidoust:


specs.json
```diff
@@ -417,6 +420,13 @@
     ]
   },
   "https://tc39.es/proposal-temporal/",
+  {
+    "url": "https://tc39.es/source-map-spec/",
```
dontcallmedom, line 424: This should replace the existing sourcemaps spec entry.

tidoust: Ah right, I had forgotten about this one...

tidoust: I propose to reuse the older entry, updating just the URL. The shortname will likely change in the future (when TC39 gives the spec an `ecma-xxx` name) in any case, and this avoids having to track the former shortname for the benefits of Specref.

dontcallmedom (approved):

PR #1049, opened 2023-09-04
Add Permissions Registry, consider registries in find-specs

tidoust:
This update also relaxes the rule on the nightly status to also accept "Draft Registry". Technically speaking, that's sort of wrong because the nightly version should rather be an "Editor's Draf of a Draft Registry", but that seems uselessly picky.

### DIFF
--- a/src/find-specs.js
+++ b/src/find-specs.js
@@ -77,7 +77,7 @@ const hasUntrackedURL = ({spec: url}) => !specs.find(s => s.nightly.url.startsWi
                                             ));
 const hasUnknownTrSpec = ({spec: url}) => !specs.find(s => s.release && trimSlash(s.release.url) === trimSlash(url)) && !specs.find(s => hasMoreRecentLevel(s,url));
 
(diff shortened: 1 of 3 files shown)


src/find-specs.js
```diff
@@ -77,6 +77,7 @@ const hasUntrackedURL = ({spec: url}) => !specs.find(s => s.nightly.url.startsWi
                                             ));
 const hasUnknownTrSpec = ({spec: url}) => !specs.find(s => s.release && trimSlash(s.release.url) === trimSlash(url)) && !specs.find(s => hasMoreRecentLevel(s,url));
 
+const composeFilters = (f1, f2) => value => f1(value) || f2(value);
```
dontcallmedom, line 80: compose is a bit generic (not clear if it is `or` or `and`) - maybe `eitherFilters` or `unionFilters`?

tidoust: True. Code now uses `eitherFilter`.

dontcallmedom (approved):

PR #1058, opened 2023-09-18
Add WasmGC spec

tomayac:
https://webassembly.github.io/spec/core/

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -452,6 +452,10 @@
     "url": "https://webassembly.github.io/exception-handling/js-api/",
     "forkOf": "wasm-js-api-2"
   },
+  {
+    "url": "https://webassembly.github.io/gc/core/",
+    "forkOf": "wasm-core" 
+  },
   "https://webbluetoothcg.github.io/web-bluetooth/",
   "https://webidl.spec.whatwg.org/",
   "https://websockets.spec.whatwg.org/",
--- a/specs.json
+++ b/specs.json
@@ -453,7 +453,7 @@
     "forkOf": "wasm-js-api-2"
   },
   {
-    "url": "https://webassembly.github.io/gc/core/",
+    "url": "https://webassembly.github.io/gc/core/bikeshed/",
     "forkOf": "wasm-core-1"
   },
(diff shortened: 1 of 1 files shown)


specs.json
```diff
@@ -452,6 +452,10 @@
     "url": "https://webassembly.github.io/exception-handling/js-api/",
     "forkOf": "wasm-js-api-2"
   },
+  {
+    "url": "https://webassembly.github.io/gc/core/",
+    "forkOf": "wasm-core"
```
tomayac, line 457:
```suggestion
    "forkOf": "wasm-core-1" 
```

specs.json
```diff
@@ -452,6 +452,10 @@
     "url": "https://webassembly.github.io/exception-handling/js-api/",
     "forkOf": "wasm-js-api-2"
   },
+  {
+    "url": "https://webassembly.github.io/gc/core/",
+    "forkOf": "wasm-core-1"
```
dontcallmedom, line 457:
```suggestion
    "forkOf": "wasm-core-1"
```
trailing white space triggers lint error

specs.json
```diff
@@ -452,6 +452,10 @@
...
```
dontcallmedom, line 457:
```suggestion
    "forkOf": "wasm-core-2"
```
it seems to be a fork of level 2 - but then for some reason, it looks like we don't have level 2 in specs.json - @tidoust, does that ring a bell?

tidoust: See https://github.com/w3c/browser-specs/pull/593 but we should revisit that. I created #1059 accordingly.

specs.json
```diff
@@ -452,6 +452,10 @@
     "url": "https://webassembly.github.io/exception-handling/js-api/",
     "forkOf": "wasm-js-api-2"
   },
+  {
+    "url": "https://webassembly.github.io/gc/core/",
```
tidoust, line 456: We're going to need that to point to the Bikeshed version, as our parsing tools cannot parse this format.

```suggestion
    "url": "https://webassembly.github.io/gc/core/bikeshed/",
```

tidoust (approved):

PR #1082, opened 2023-10-10
Some updates based on #1008

tidoust:
- Add WebXR Plane Detection
- Add WebXR Meshing API
- Add WebXR Mesh Detection Module
- Add WebBluetooth Scanning API
- Add Managed Configuration API
- Add MathML4, as non browser spec
- Monitor Private Aggregation API (pending w3c.json at least)
- Ignore a few additional specs (not required as these specs are not found but that makes the rationale explicit)

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -153,6 +153,14 @@
   "https://immersive-web.github.io/anchors/",
   "https://immersive-web.github.io/model-element/",
   "https://immersive-web.github.io/raw-camera-access/",
+  {
+    "url": "https://immersive-web.github.io/real-world-geometry/plane-detection.html",
+    "shortname": "webxr-plane-detection"
+  },
+  {
+    "url": "https://immersive-web.github.io/real-world-geometry/webxrmeshing-1.html",
+    "shortname": "webxr-meshing"
+  },
   "https://immersive-web.github.io/real-world-meshing/",
   "https://infra.spec.whatwg.org/",
   "https://mimesniff.spec.whatwg.org/",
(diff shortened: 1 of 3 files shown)


specs.json
```diff
@@ -153,6 +153,14 @@
   "https://immersive-web.github.io/anchors/",
   "https://immersive-web.github.io/model-element/",
   "https://immersive-web.github.io/raw-camera-access/",
+  {
+    "url": "https://immersive-web.github.io/real-world-geometry/plane-detection.html",
```
dontcallmedom, line 157: The IDL of that spec is incomplete (missing `[Exposed]` on the interface), which IIUC will block the IDL package release down the line. I'm curious if there is a better way to deal with this than just going through the cascade of releases.

tidoust: I do not have a fully satisfactory solution to propose today, either:
1. we put the spec in the monitor list for now since we know that the IDL is invalid; or
2. we create a patch; or
3. we wait until the IDL gets fixed

I'm surprised that we did not seem to have any open issue on the topic, even though we've been floating the idea to improve the curation process in that regard for some time. I created https://github.com/w3c/webref/issues/1055 to track this.

Another option could be to automate checks a bit more when we add a spec to browser-specs, e.g., by running Reffy and tests on the spec and prevent addition of the spec while checks fail. That seems nice in theory but more complex in practice. Also, that wouldn't solve the problem for specs that start to fail after some time.

dontcallmedom (approved):
thanks! let's proceed with this and we'll deal with the consequences as they emerge this time around; we now have w3c/webref#1055 to look into a more robust approach

PR #1090, opened 2023-10-12
Explicitly ignore execCommand and sourcing in-band track

tidoust:
Discussed in #1008. Both specs are referenced by HTML. Both specs are in a sorrow state with no plan to resume work in the foreseeable future. Adding them to the ignore list as a result.

We may want to revisit that if the situation evolves, but I don't think we should (or want ;)) to do that every 3 months.

### DIFF
--- a/src/data/ignore.json
+++ b/src/data/ignore.json
@@ -423,12 +423,6 @@
     },
     "https://w3c.github.io/math-aam/": {
       "comment": "no longer worked on and community group closed"
-    },
-    "https://w3c.github.io/editing/docs/execCommand/": {
-      "comment": "no longer worked on and not implemented consistently or fully by user agents"
-    },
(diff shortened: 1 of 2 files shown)


dontcallmedom:
an alternative may to be put them in monitor but with a made up "last-reviewed" date somewhat far in the future? I'm a bit worried that putting them in the ignore list would hide them even longer than we would want.

tidoust:
Which makes me realize: do we actually need to bother with a date set in the future? I thought we were reviewing all specs in the monitor list every 3 months, but we only review those that have changed by the last time we reviewed them, right? I don't think that these specs will pass that criterion until something happens.

dontcallmedom:
even better indeed :)

dontcallmedom (approved):

PR #1099, opened 2023-10-19
Update URL of urlpattern spec following move to WHATWG

tidoust:
Move is ongoing and resulting spec is not in place yet, but then previous spec is no longer in place either...

### DIFF
--- a/src/data/ignore.json
+++ b/src/data/ignore.json
@@ -90,7 +90,7 @@
       "comment": "not targeted at browsers"
     },
     "WICG/urlpattern": {
-      "comment": "merged into HTML"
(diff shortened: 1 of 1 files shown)


src/data/ignore.json
```diff
@@ -89,6 +89,9 @@
     "WICG/open-ui": {
       "comment": "not targeted at browsers"
     },
+    "WICG/urlpattern": {
+      "comment": "merged into HTML"
```
dontcallmedom, line 93:
```suggestion
      "comment": "moved to WHATWG"
```

dontcallmedom (approved):

PR #1129, opened 2023-11-17
[README] Describe spec ids and algo to track a spec over time

tidoust:
This adds a "Spec identifiers" section to the README to explain that there are no fully stable identifiers in browser-specs, but that the `shortname` and `formerNames` properties may be used to track a spec over time if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ cross-references, WebIDL, quality, etc.
     - [`tests.testPaths`](#teststestpaths)
     - [`tests.excludePaths`](#testsexcludepaths)
   - [`source`](#source)
+- [Spec identifiers](#spec-identifiers)
 - [How to add/update/delete a spec](#how-to-addupdatedelete-a-spec)
 - [Versioning](#versioning)<!-- COMMON-TOC: end -->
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -605,6 +615,53 @@ The provenance for the `title` and `nightly` property values. Can be one of:
...
+Group, or when a new version of an IETF draft gets published.
+
+If your project tracks specifications over time and relies on browser-specs to
+gather information about these specifications, you will need to record the
+`shortname` of the specifications you're tracking, and apply the following
+algorithm to find the relevant specification entry in browser-specs:
+
+1. Look for an entry in browser-specs whose `shortname` matches the recorded
+shortname. If one is found, that is the relevant specification entry.
+2. Look for the first entry in browser-specs that has the recorded shortname in
```
dontcallmedom, line 653: should we add a test that no two specs should overlapping formerNames? or is that not something we guarantee either?

tidoust: I hesitated but that seems something that we can at least guarantee for the foreseeable future. I'm adding a test and updating the algorithm to note the guarantee.

PR #1135, opened 2023-11-21
Automate processing of IETF (group) drafts

### LINKED ISSUE #1122: Learn to parse IETF datatracker pages
IETF entries that have not yet been published as RFCs have a canonical URL that looks like `https://datatracker.ietf.org/doc/html/...`.

Such entries need to be explicit about the organization, the group, and more often than not, a "better looking" nightly URL, e.g. one under `https://www.ietf.org/archive/`. The problem with the nightly URL is that it typically contains the current revision of the draft and thus becomes outdated as soon as a new revision is published.

As we add more of these specs, the code could rather:
1. know that it's an IETF spec
2. fetch the datatracker page, which can be directly derived from the canonical URL such as https://datatracker.ietf.org/doc/draft-zern-webp/ to get the latest nightly URL.

Not sure yet how to extract the group itself though.

tidoust:
The code now recognizes IETF draft documents that have a `datatracker.ietf.org` URL:
- It associates them with the IETF organization
- It can compute a useful shortname (that code can in theory return a truncated shortname because there is no direct way to validate that the Internet Draft name contains a group ID).
- It extracts the group's ID from the nightly URL (that code could further be improved to fetch the actual group name, right now the code only knows about the "HTTP" working group).
- It associates IETF documents from the HTTP WG to the right repository.
- It computes the better-looking nightly URL at `www.ietf.org` or at `httpwg.org` for HTTP WG documents.

This allows to simplify IETF data in `specs.json` a bit.

Note that the code still cannot process drafts that have been submitted by individuals automatically, even when these drafts at targeted at a group. Such drafts should be associated with the individuals that submitted them and not with any group. A couple of spec entries, which incorrectly referenced the Network WG or the HTTP WG, were fixed accordingly in `specs.json`.

This fixes #1122, but note that the code does not need to fetch the datatracker page for the time being.

### DIFF
--- a/README.md
+++ b/README.md
@@ -610,6 +610,7 @@ The `excludePaths` property is seldom set.
 The provenance for the `title` and `nightly` property values. Can be one of:
 - `w3c`: information retrieved from the [W3C API](https://w3c.github.io/w3c-api/)
 - `specref`: information retrieved from [Specref](https://www.specref.org/)
+- `ietf`: information retrieved from the [IETF datatracker](https://datatracker.ietf.org)
 - `spec`: information retrieved from the spec itself
 
 The `source` property is always set.
--- a/schema/definitions.json
+++ b/schema/definitions.json
@@ -60,7 +60,7 @@
 
     "source": {
       "type": "string",
-      "enum": ["w3c", "specref", "spec"]
+      "enum": ["w3c", "specref", "spec", "ietf"]
     },
 
     "release": {
--- a/specs.json
+++ b/specs.json
@@ -18,12 +18,10 @@
   "https://console.spec.whatwg.org/",
   {
     "url": "https://datatracker.ietf.org/doc/html/draft-cutler-httpbis-partitioned-cookies",
-    "shortname": "partitioned-cookies",
-    "organization": "IETF",
     "groups": [
       {
-        "name": "HTTP Working Group",
-        "url": "https://datatracker.ietf.org/wg/httpbis/"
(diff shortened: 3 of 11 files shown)


dontcallmedom (approved):
This looks good to me, but while researching this PR, I discovered the datatracker API that I think would make a lot of this more reliable:
https://datatracker.ietf.org/api/

See e.g. https://datatracker.ietf.org/api/v1/doc/document/?name=draft-ietf-httpbis-digest-headers&format=json https://datatracker.ietf.org/api/v1/group/group/1718/?format=json

tidoust:
Ah, I was looking into datatracker docs and was about to make the same point :)

Perhaps even simpler, there is a simplified JSON file that contains all the information we need and that would be more readily ingestable (if we use the API, we'll have to hop through a few endpoints to collect the info), e.g.:

https://datatracker.ietf.org/doc/draft-zern-webp/doc.json
https://datatracker.ietf.org/doc/rfc6797/doc.json

Also, I don't really understand how to get information about a document published as an RFC with the main API document endpoint, e.g. https://datatracker.ietf.org/api/v1/doc/document/?name=rfc6797&format=json does not work. It seems that one has to know the draft name instead, as in https://datatracker.ietf.org/api/v1/doc/document/?name=draft-ietf-websec-strict-transport-sec&format=json and the easiest way to map the rfc to the draft name would be to fetch the related `doc.json`.

dontcallmedom (approved):

PR #1142, opened 2023-11-29
Mark replacement of DOM-Style by cssom-1

dontcallmedom:


tidoust (approved):
I wish the specs themselves would say something about it...

dontcallmedom:
cssom1 lists DOM-Style-2 in its "previous" version links fwiw

PR #1146, opened 2023-11-30
Fix datatracker alternate URL test for IETF specs

tidoust:
The datatracker URL does not have a trailing `/`.


dontcallmedom (commented):
arg, thanks for the fix and sorry for the bug in the first place :)

PR #1149, opened 2023-11-30
Fix alternate URLs computation

tidoust:
The `computeAlternateUrls` function is supposed to return the list of alternate URLs. For CSS specs, it both returned an alternate URL *and* set another one in the spec directly. Surprisingly, this did not have any bad consequence other that swapping the entries...


dontcallmedom (approved):
copy & paste-- - thanks!

PR #1150, opened 2023-11-30
Fix alternateUrls logic for discontinued specs, add related tests

tidoust:
Index building now skips over some quirks when processing a discontinued spec for which we reuse information from the previously published version.

New tests make sure that we never list duplicate alternate URLs, and that alternate URLs are *alternate* URLs indeed.


tidoust:
CI test failure actually proves that the test works as expected ;)

PR #1164, opened 2023-12-26
Flag custom-state-pseudo-class as discontinued

tidoust:
The WICG spec moved to actual standardization with two parts:

1. The IDL part was moved to HTML in https://github.com/whatwg/html/pull/8467
2. The CSS part is to be integrated in Selectors, see https://github.com/w3c/csswg-drafts/issues/4805

This flags the spec as obsoleted by these two specs, using `selectors-4` for the CSS part. The CSS WG resolution mentions Selectors 5 but it does not exist yet. Alternatively, we could keep this PR open until the CSS part gets specified, or not mention `selectors-4` at all for now. Not sure which is better. Merging earlier rather than later would help with Webref (to remove the duplicate IDL).

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -520,8 +520,7 @@
     "url": "https://wicg.github.io/custom-state-pseudo-class/",
     "standing": "discontinued",
     "obsoletedBy": [
-      "html",
-      "selectors-4"
+      "html"
     ]
   },
   "https://wicg.github.io/datacue/",
--- a/specs.json
+++ b/specs.json
@@ -516,7 +516,14 @@
     }
   },
   "https://wicg.github.io/css-parser-api/",
-  "https://wicg.github.io/custom-state-pseudo-class/",
+  {
+    "url": "https://wicg.github.io/custom-state-pseudo-class/",
+    "standing": "discontinued",
+    "obsoletedBy": [
+      "html",
+      "selectors-4"
+    ]
+  },
   "https://wicg.github.io/datacue/",
   "https://wicg.github.io/deprecation-reporting/",
   "https://wicg.github.io/digital-goods/",


specs.json
```diff
@@ -516,7 +516,14 @@
     }
   },
   "https://wicg.github.io/css-parser-api/",
-  "https://wicg.github.io/custom-state-pseudo-class/",
+  {
+    "url": "https://wicg.github.io/custom-state-pseudo-class/",
+    "standing": "discontinued",
+    "obsoletedBy": [
+      "html",
+      "selectors-4"
```
dontcallmedom, line 524:
```suggestion
      "selectors-5"
```
As annoying as it is, I think the right solution is not to merge this until the CSS bits actually surface in a new css-selectors-5; at least, here is my reasoning:
* neither the spec nor the repo are marked as discontinued
* clearly the HTML spec has taken over the IDL bit, but this is properly taken care of by your patch in https://github.com/w3c/webref/pull/1119/files
* you can't quite make sense of the feature without what's described in the spec until it gets [moved into the new CSS selector 5 spec ](https://github.com/w3c/csswg-drafts/pull/8213/files)
* referencing css-selectors-4 when it doesn't actually describe the feature doesn't really help anyone

tidoust: Works for me. Let's park this for now!

dontcallmedom:
https://wicg.github.io/custom-state-pseudo-class/ now redirects to the HTML spec, which breaks the webref patch.

Given there is no sign of imminent progress on css-selectors-5, I guess we should make the spec obsoletedby html only at this point, and file an issue to add the new selectors spec when it arises. I'll update the PR in that direction.

Issue #1059, opened 2023-09-18
See what changes are needed to support WASM Core Level 2


tidoust:
Via https://github.com/w3c/browser-specs/pull/593 and https://github.com/w3c/browser-specs/pull/1058#pullrequestreview-1630650568

dontcallmedom:
isn't it a matter of using https://webassembly.github.io/gc/core/bikeshed/ as crawled url?

tidoust:
For the crawl of nightly versions, yes. However, the [release version](https://www.w3.org/TR/wasm-core-2/) is not the Bikeshed version. I don't remember whether that creates a real problem during the crawl or just means that the release crawl creates mostly empty extracts. That's what I'd like to look into.

PR #1190, opened 2024-02-05
Add WASM Core 2 spec and phase 3+ WASM proposals

tidoust:
The list only contained Level 1 of WASM Core (see #1059). This adds Level 2. In practice, crawling will work fine for the Editor's Draft but Reffy will basically not extract anything from the /TR version of the spec because it does not follow usual patterns. That seems fine enough for now.

This update also introduces missing WASM proposals as forks of the WASM Core or WASM JavaScript API specification. Most of the time, the title needs to be created because the actual specification remains that of the base spec.

One difficulty is that the WebAssembly group approaches extensions proposals as generic WASM extensions, and not necessarily as WASM *Core* extensions or WASM *JS API* extensions, whereas we need to make a choice to set the `forkOf` property. Things look good for this batch of updates, because current proposals still seem to extend either of these specs. That may not always be the case though in the future!

Similarly, the find-specs script assumed that proposals were always extending the Core spec. It now reports the URL of the home page, both to make it clearer that a choice needs to be made, and to avoid reporting a proposal that is already in the list as a WASM JS API fork.

Via #1186.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -473,12 +473,63 @@
   "https://w3c.github.io/webdriver-bidi/",
   "https://w3c.github.io/webrtc-ice/",
   {
+    "url": "https://webassembly.github.io/branch-hinting/core/bikeshed/",
+    "forkOf": "wasm-core-2",
+    "title": "WebAssembly Core: Branch Hinting"
+  },
+  {
+    "url": "https://webassembly.github.io/content-security-policy/js-api/",
+    "forkOf": "wasm-js-api-2",
+    "title": "WebAssembly JavaScript Interface: Content Security Policy"
+  },
+  {
     "url": "https://webassembly.github.io/exception-handling/js-api/",
     "forkOf": "wasm-js-api-2"
   },
   {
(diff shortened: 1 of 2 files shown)


specs.json
```diff
@@ -472,13 +472,64 @@
...
+    "title": "WebAssembly Core: Branch Hinting"
+  },
+  {
+    "url": "https://webassembly.github.io/content-security-policy/js-api/",
+    "forkOf": "wasm-js-api-2",
+    "title": "WebAssembly JavaScript Interface: Content Security Policy"
+  },
   {
     "url": "https://webassembly.github.io/exception-handling/js-api/",
     "forkOf": "wasm-js-api-2"
```
tidoust, line 487: Note this one has a custom title, no need to set it ourselves in `specs.json`.

dontcallmedom (approved):
I note that the forks include forks of the test suites, which may or may not include related test cases; I think we're probably better off ignoring this at the moment

Issue #1191, opened 2024-02-05
How to add ISO/IEC specs


tidoust:
Via #1089. ISO/IEC standards such as [JPEG XL](https://www.iso.org/standard/85253.html) do not have a public specification. They all have a "canonical" page on the ISO web site though, which seems a good entry point for the spec. That URL is also the one [used in Specref](https://www.specref.org/?q=jpeg%20xl).

Now, all entries in browser-specs so far target actual specs. Question is: how should we add ISO/IEC entries to browser-specs?

A clean solution would probably be to use the ISO page URL as the canonical URL and **not to set** the `nightly` property, because there are no public spec to point at. That would break the schema though as `nightly` is mandatory, and would likely require updates to the code of projects that currently consume browser-specs data as these projects may expect `nightly.url` to be always set.

A non-breaking solution would be to use the ISO page URL as the canonical URL and as the `nightly.url` URL. That's less clean but most code that uses browser-specs data today should continue to run. Well, Reffy may still need to learn to skip `iso.org` URLs.

Overall, choice seems to be between a major bump and a minor bump. The major bump approach is more painful but has the advantage of making things explicit. The minor bump is easier but has the drawback of hiding the fact that we may not have a published version of the spec.

dontcallmedom:
I prefer the major bump approach ("the future is longer than the past", he said optimistically)

tidoust:
Corresponding updates to code and list made, and new major versions released (already deployed in web-features and browser-compat-data, [pending PR for Yari](https://github.com/mdn/yari/pull/10478)). I'll update Reffy accordingly.

Issue #1089, opened 2023-10-12
Consider web font/image/media formats


tidoust:
Via #1008. Consider adding font/image/media formats commonly supported on the web. The main difficulty is that these specifications do not follow a well-known format.

The list includes:
- [x] GIF
- [x] JPEG
- [x] [WebP Container Specification](https://developers.google.com/speed/webp/docs/riff_container)
- [x] [WebP Lossless Bitstream](https://developers.google.com/speed/webp/docs/webp_lossless_bitstream_specification)
- [x] [COLR — Color Table](https://learn.microsoft.com/en-us/typography/opentype/otspec191alpha/colr)
- [x] GLSL: Shading language for WebGL.
- [x] glTF: File format for 3D models
- Various more image, video, audio, font, etc. formats.

tidoust:
For WebP, see dedicated issue #1109

foolip:
https://github.com/web-platform-dx/web-features/pull/580 is blocked on this, and is additionally made complicated by the fact that the JEPG XL spec doesn't have a public URL.

tidoust:
> [web-platform-dx/web-features#580](https://github.com/web-platform-dx/web-features/pull/580) is blocked on this, and is additionally made complicated by the fact that the JEPG XL spec doesn't have a public URL.

I think we can take the same approach as [Specref](https://www.specref.org/?q=jpeg%20xl), which links to the spec page on the ISO web site for ISO/IEC standards. For JPEG XL support in browsers, we probably want to document the file format: [iso18181-2](https://www.iso.org/standard/85253.html), not the core coding system (or list both, but the file format seems the right entry point). I created #1191 to discuss adding specs that don't have published versions to browser-specs.

([Can I Use](https://caniuse.com/jpegxl) rather links to the [JPEG Documentation page](https://jpeg.org/jpegxl/documentation.html), which seems suboptimal: the page does not linnk to the ISO web site)

tidoust:
The list now contains the formats identified in this issue. A couple of notes:

- For JPEG, the reference is to [`iso10918-5`](https://www.iso.org/standard/54989.html), done in #1237
- For GLSL, the spec actually added is [OpenGL ES Shading Language](https://registry.khronos.org/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.html) (ESSL) because I believe that's what WebGL actually uses, done in #1238
- For COLR, [`iso14496-22`](https://www.iso.org/standard/87621.html) was used, done in #1211

[...]

PR #1192, opened 2024-02-05
[breaking] Add support for ISO standards, add JPEG XL ISO/IEC spec

tidoust:
To add ISO/IEC standards related to JPEG XL (see #1089), there needs to be a way to add an entry in browser-specs that has a canonical URL but no actual nightly URL, because ISO standards are not public (see also #1191).

This update amends the code to allow and create entries without a `nightly` property when needed (only for ISO standards for now). The code also retrieves the name of the group that develops an ISO standard.

This is a BREAKING CHANGE because the `nightly` property used to be mandatory. Projects that expect to find a `nightly.url` property need to be updated to only look at the root `url` property or to skip the entry altogether.

First ISO spec added to the list is JPEG XL, which will appear as:

```json
{
  "url": "https://www.iso.org/standard/85253.html",
  "seriesComposition": "full",
  "shortname": "iso18181-2",
  "series": {
    "shortname": "iso18181-2",
    "currentSpecification": "iso18181-2",
    "title": "Information technology — JPEG XL image coding system — Part 2: File format",
    "shortTitle": "JPEG XL: File Format"
  },
  "shortTitle": "JPEG XL: File Format",
  "organization": "ISO/IEC",
  "groups": [
    {
      "name": "ISO/IEC JTC 1/SC 29",
      "url": "https://www.iso.org/committee/45316.html"
    }
  ],
  "title": "Information technology — JPEG XL image coding system — Part 2: File format",
  "source": "specref",
  "categories": [
    "browser"
  ],
  "standing": "good"
}
```

Worth noting: the absence of a `nightly` property means that there is no place to store the status of the spec, which could in theory be "Under development" or "Published" for ISO specs (there are additional stages in the ISO process but they are probably not worth capturing in any case). An alternative would be to have a `nightly.status` property, but that seems clumsy.

### DIFF
--- a/src/fetch-groups.js
+++ b/src/fetch-groups.js
@@ -11,6 +11,62 @@ const parseSpecUrl = require("./parse-spec-url.js");
 
 
 /**
+ * Retrieve the information about the exact organization and the group that
+ * develops an ISO specification from the description page on the ISO web site.
+ *
+ * The group's name and URL appear in the page under the General Information
+ * heading.
+ *
+ * Note: It would be better to use Puppeteer to parse the HTML, but that seems
+ * a bit overkill (and would not be future proof either since the HTML page
+ * does not contain good non-text anchors to extract the info we need.
+ */
+async function setISOGroupFromPage(spec, options) {
+  const res = await fetch(spec.url, options);
+  if (res.status !== 200) {
+    throw new Error(`Could not retrieve ISO page ${spec.url}, status code is ${res.status}`);
+  }
+  const html = await res.text();
+  let startPos = html.indexOf('<h3>General information</h3>');
+  if (startPos === -1) {
+    throw new Error(`Cannot find General information heading in ISO page ${spec.url}`);
+  }
+  startPos = html.indexOf('Technical Committee&nbsp;:', startPos);
+  if (startPos === -1) {
+    throw new Error(`Cannot find technical committee information in ISO page ${spec.url}`);
+  }
+  startPos = html.indexOf('<a ', startPos);
+  if (startPos === -1) {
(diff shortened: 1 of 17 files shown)


test/fetch-info.js
```diff
@@ -216,6 +216,18 @@ describe("fetch-info module", function () {
...
+
+    it("extracts spec info from an ISO spec page", async () => {
+      const spec = {
+        url: "https://www.iso.org/standard/61292.html",
+        shortname: "iso18074-2015"
+      };
+      const info = await fetchInfo([spec]);
+      assert.ok(info[spec.shortname]);
+      assert.equal(info[spec.shortname].source, "spec");
+      assert.equal(info[spec.shortname].title, "Textiles — Identification of some animal fibres by DNA analysis method — Cashmere, wool, yak and their blends");
```
dontcallmedom, line 228: 😂

src/fetch-groups.js
```diff
@@ -10,6 +10,62 @@ const Octokit = require("./octokit");
...
+ * a bit overkill (and would not be future proof either since the HTML page
+ * does not contain good non-text anchors to extract the info we need.
+ */
+async function setISOGroupFromPage(spec, options) {
+  const res = await fetch(spec.url, options);
+  if (res.status !== 200) {
+    throw new Error(`Could not retrieve ISO page ${spec.url}, status code is ${res.status}`);
+  }
+  const html = await res.text();
+  let startPos = html.indexOf('<h3>General information</h3>');
```
dontcallmedom, line 30: 😭

dontcallmedom (approved):

PR #1204, opened 2024-02-12
Add arraybuffer-base64 and digital identities specs 

dontcallmedom:


tidoust (changes requested):
I don't like the idea of using the presence of `null` as a signal. First thing I want to do when I see a property with a null value is to remove it as useless.

We already handle a few exceptions in [`src/compute-shortname.js`](https://github.com/w3c/browser-specs/blob/main/src/compute-shortname.js#L175), including for published ECMAScript specs. I would rather extend that list to TC39 proposals.

Alternatively, a more generic mechanism could be to add an explicit `delete` (or `skip`) property in `specs.json`, something like:

```json
{
  "url": "https://tc39.es/proposal-arraybuffer-base64/spec/",
  "delete": ["seriesVersion"]
}
```

src/compute-shortname.js
```diff
@@ -179,6 +179,7 @@ function completeWithSeriesAndLevel(shortname, url, forkOf) {
   // digits, as in "iso18181-2".
   if (seriesBasename.match(/^ecma-/) ||
       seriesBasename.startsWith("iso") ||
+      seriesBasename.endsWith("base64") ||
```
tidoust, line 182: Hmm, I was thinking of a more generic "no series version for TC39 proposals" exception, since in practice these proposals do not, and should not, contain any version number.

Looking into it, the exception could actually be "no series version for TC39 specs" since, at least for now, we don't try to keep track of versions of TC39 specs:

```suggestion
      url.match(/^https:\/\/tc39\.es\//) ||
```

This also allows to drop the first `seriesBasename.match(/^ecma-/)` exception as underlying URLs also start with `https://tc39.es/`.

dontcallmedom: much more elegant, indeed :) will fix

Maybe the generalization of this is using an allowlist for organizations that do use series version? (or a blocklist of the reverse); that's probably not practical at compute-shortname stage, but could be done in a clean up phase?

tidoust: Ah well, the approach based on the URL is not going to work as-is, due to the fact that shortname was made explicit, which effectively replaces the URL in the call to `computeShortname`. See #1219.

I'll update to a check on the shortname instead.

PR #1210, opened 2024-02-15
Add IntersectionObserver v2

foolip:


dontcallmedom (changes requested):
we already have https://www.w3.org/TR/intersection-observer/ in the list, which points to https://w3c.github.io/IntersectionObserver/ which seems more recently updated than "https://w3c.github.io/IntersectionObserver/v2/"

foolip:
https://w3c.github.io/IntersectionObserver/v2/ includes things that aren't in v1, namely the `trackVisibility` parameter for the `IntersectionObserver` constructor and the `isVisible` attribute on `IntersectionObserverEntry`.

I'm adding this because I need it for https://github.com/web-platform-dx/web-features/pull/620, AFAIK this is the only spec that describes the additions.

@szager as the listed editor of the v2 spec, do you have thoughts?

dontcallmedom:
see also https://github.com/w3c/IntersectionObserver/issues/468 from a certain @foolip :)

dontcallmedom:
with https://github.com/w3c/IntersectionObserver/pull/523 I think there won't be a need to track a v2 separately, so closing

PR #1214, opened 2024-02-19
Adjust test on IETF draft nightly URL extraction

tidoust:
Previous IETF draft used in the test was published as an RFC, and is no longer suitable for the test.

More generally speaking, we cannot expect these tests to remain completely stable over time unless we start mocking the response from the server... but then it's also good to check that the response we get from the server is as expected. In the meantime, I propose to consider that these tests are stable enough!


dontcallmedom (approved):
Rah, I couldn't make sense of why the test was failing with updated data - I hadn't thought of checking the test data >:)

Issue #1220, opened 2024-02-19
Consider rules per organization


tidoust:
Via https://github.com/w3c/browser-specs/pull/1204#discussion_r1494559370.

Some rules, currently hardcoded as exceptions to the rule, could perhaps be expressed through more declarative allowlists/blocklists per organization. This starts with `computeShortname` but can perhaps be extended to other modules.

PR #1239, opened 2024-02-29
Add Khronos Group specs to the list of exceptions for repository test

tidoust:
New Khronos Group specs do not have a repository. The test now stops looking at Khronos Group specs. That's a bit loose since all WebGL specs do link to a repository, but that test is not meant to be too precise either.


tidoust:
Also worth improving when #1220 gets considered.

PR #1255, opened 2024-03-11
Adjust URL of long-animation-frame

tidoust:
Meanwhile in Veracruz, the repository was renamed and the shortname changed.

I suspect the title will be updated soon to reflect the switch to singular as well.

Cc @tunetheweb

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -470,7 +470,7 @@
   "https://w3c.github.io/audio-session/",
   "https://w3c.github.io/contentEditable/",
   "https://w3c.github.io/gamepad/extensions.html",
-  "https://w3c.github.io/long-animation-frames/",
+  "https://w3c.github.io/long-animation-frame/",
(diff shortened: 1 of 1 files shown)


tunetheweb:
Do you know why this changed @noamr ? Would have thought it should have a trailing "s" to be consistent with https://w3c.github.io/longtasks/

noamr:
> Do you know why this changed @noamr ? Would have thought it should have a trailing "s" to be consistent with https://w3c.github.io/longtasks/

Yes, sorry, I asked @caribouW3 to make an adjustment and forgot the `s` in the email. Let's correct this.

tunetheweb:
And by "correct this" you mean to put the "s" back, so this PR here should not be needed then?

noamr:
> And by "correct this" you mean to put the "s" back, so this PR here should not be needed then?

YEs.

PR #1254, opened 2024-03-11
Add Long Animation Frames (LoAF)

tunetheweb:
Shipping in Chrome 123 and now separated from the Long Tasks spec: https://github.com/w3c/longtasks/issues/132#event-12069245858

PR #1257, opened 2024-03-13
Update monitor.json to add long-animation-frames

noamr:


tidoust:
A spec can either be in the main list or in the monitor list, and `long-animation-frames` was added to the main list a couple of days ago through #1254 (and included in the packaged version of web-specs released earlier today). It shouldn't be added to the monitor list! Or did you have something specific in mind?

noamr:
Oops, double work (first time I'm doing this, sorry)

tidoust:
Well, nothing explains what the monitor list is in practice. I created #1258 to improve the documentation so that we can at least point people at something to read ;)

PR #1261, opened 2024-03-15
Also re-generate specs in the same series when building diff

tidoust:
So far, the build-diff tool only generated specs that effectively changed in `specs.json`. With this change, it now also re-generate the entries for the specs that are in the same series, and report possible changes to these specs in a new `seriesUpdate` property.

The change also switches to active forms for the property names: `add`, `update` and `delete` (instead of `added`, `updated`, `deleted`), because the changes to `index.json` are typically not yet done.

This change gets us closer to a mode where we can pre-compute the `index.json` that a pull request would trigger and run tests on them.

Still missing: forks. If changes add a `forkOf` property, include a rename of a spec that has a `forkOf` property, or delete a fork, the upstream spec entry needs to be re-generated as well. The code still does not handle that.

### DIFF
--- a/src/build-diff.js
+++ b/src/build-diff.js
@@ -1,9 +1,75 @@
+const assert = require("assert");
 const path = require("path");
 const { execSync } = require("child_process");
 const { generateIndex } = require("./build-index");
+const computeShortname = require("./compute-shortname.js");
 
+/**
+ * Spec sort function.
+ *
+ * Lists are sorted by ascending URL.
+ */
 function compareSpecs(a, b) {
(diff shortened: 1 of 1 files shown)


tidoust:
> Still missing: forks. If changes add a forkOf property, include a rename of a spec that has a forkOf property, or delete a fork, the upstream spec entry needs to be re-generated as well. The code still does not handle that.

Well, good news is that this "still missing" note is wrong ;) The script already supports that: a fork is in the same series as the specification it is a fork of!

PR #1264, opened 2024-03-19
Update build-diff script to also accept a URL

tidoust:
This update makes the `src/build-diff.js` script accept the canonical URL of a spec as parameter, and build the updates to `index.json` that this URL would trigger if it were added to `specs.json` (or if it would replace the corresponding entry in `specs.json`).

The script still accepts commit references, although the syntax was slightly updated to get back to one parameter and better align with `git diff`, from `node src/build-diff HEAD HEAD~2` to `node src/build-diff HEAD..HEAD~2`.

The script is now also available as a module and the exported function also accepts additional spec properties on top of the URL. That is not yet used but will be useful for later.

Next step is to create a more proper CLI that makes use of the module, can also run tests on the updates, and propose a human-readable report that could be added to a pull request and/or used as commit message.

### DIFF
--- a/src/build-diff.js
+++ b/src/build-diff.js
@@ -206,7 +206,7 @@ async function buildCommits(newRef, baseRef, { diffType = "diff", log = console.
  * The spec object must have a `url` property. It may have other properties.
  *
  * Internally, the function turns the parameter into a diff and hands it over
- * to `buildSpec`.
+ * to `buildDiff`.
  *
  * The function throws in case of errors.
  */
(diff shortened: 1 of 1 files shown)


src/build-diff.js
```diff
@@ -119,6 +196,45 @@ async function compareIndex(newRef, baseRef, { diffType = "diff", log = console.
...
+}
+
+
+/**
+ * Build the diff for the given spec.
+ *
+ * The spec object must have a `url` property. It may have other properties.
+ *
+ * Internally, the function turns the parameter into a diff and hands it over
+ * to `buildSpec`.
```
dontcallmedom, line 209:
```suggestion
 * to `buildDiff`.
```

dontcallmedom (approved):

PR #1270, opened 2024-03-25
Add "browser-specs" CLI and issue workflow to manage lists

tidoust:
*Draft PR as I still need to test the template and jobs, which probably contain bugs. This should be good enough for a review though, in particular to discuss potential changes to the suggested workflow. One potential change that comes to mind would be to say: "screw the PR step, let's push to "main" directly, because it's not clear what added value the PR would bring, on top of introducing another review step.*

This adds a new CLI, to be run with `npx browser-specs`. The CLI wraps `build-diff.js` and eases the insertion of individual specs to the lists (`specs.json`, `monitor.json` and `ignore.json`). In particular, the command:

- reports changes that would be made to `index.json`;
- runs tests against these changes;
- takes care of linting;
- removes entries from the monitor/ignore list if needed;
- can commit changes and create a pull request out of them.

See inline command help for details.

The goal is to have a command that is both useful for local execution and for processing issues on GitHub. The CLI reports markdown for that reason.

For a local run, the build command takes a URL as parameter (unless the goal is to test changes already made to `specs.json`). This should be all that's needed to add a spec in about 90% of cases. Additional properties can be specified through the `--json` option.

The build command may also take an issue number as parameter, in which case it will retrieve the spec information from GitHub. This requires the `gh` CLI to be available. This is intended to be used along with an issue template that sets the structure of a spec suggestion issue.

A new issue template allows to collect suggestions to add a new spec in a structured issue. As for the command-line, the issue template essentially asks: "what is the spec URL?", and leaves the possibility to specify a JSON object to define additional properties. That is not super friendly but should not be useful in the vast majority of cases. Issue creation triggers a bot that runs the CLI and reports the result in an issue comment. The same issue comment is reused if the bot runs multiple times on the issue (e.g., because the issue was edited).

Once repo maintainers are satisfied with the result, they can add an `@browser-specs-bot pr` to trigger the creation of PR out of the issue.

NB: The actions require the creation of a `BOT_TOKEN` secret associated with a `@browser-specs-bot` account to run the `gh` CLI command with appropriate rights.

This is a first stab. The code could benefit from refactoring here and there, and I'm sure that the code/workflow will need to be adjusted.

Next steps include:
- Describe the mechanism in the README and CONTRIBUTING files.
- Use the CLI to test pull requests (for when people suggest a spec through a PR).
- Use this mechanism to propose new specs reported by `find-specs.js`
- Add mechanism to say something like `@browser-specs-bot monitor Too early` in a spec suggestion issue to add the spec to the monitor/ignore lists.


src/cli.js
```diff
@@ -0,0 +1,547 @@
...
+  else {
+    report += 'These changes look good! 😎\n';
+  }
+
+  return report;
+}
+
+
+/**
+ * Helper function to split a session issue body (in markdown) into sections
```
dontcallmedom, line 126:
```suggestion
 * Helper function to split an issue body (in markdown) into sections
```

src/cli.js
```diff
@@ -0,0 +1,547 @@
...
+  .argument('<what>', 'what to build. The argument may either be a `<url>` that represents the canonical URL of the spec to build, a named commit `<commit>` to point at the git commit that contains the changes to build in `specs.json`, two arbitrary named commits `<commit>..<commit>` to compile the list of changes made to `specs.json` between the two commits, or an issue number of the w3c/browser-specs repo that represents a spec suggestion and follows the expected structure. The `gh` CLI command must be available if `<what>` is an issue number.')
+  .option('-c, --commit', 'commit potential updates to a dedicated branch and switch to that branch. This option implies the `--update` option. The `git` CLI command must be available.')
+  .option('-j, --json <json>', 'link to a JSON file with additional spec properties, or a serialized JSON object directly. The option can only be set if `<what>` is a URL. The option value can also be the JSON object directly enclosed in double quotes (or without whitespaces), e.g. `"{ ""nightly"": { ""sourcePath"": ""compatibility.bs"" } }"`. A `url` property at the root level will be ignored.')
```
dontcallmedom, line 160: shouldn't this be enclosed in single quotes instead? I'm not sure how `"{ ""nightly"": { ""sourcePath"": ""compatibility.bs"" } }"` can be parsed into valid JSON

dontcallmedom (approved):
Oof, impressive and looking very appealing!

Since the code is already written for it, I suggest we try the go-through-pr mode as a starting point, and consider skipping it once we're confident we can YOLO from the issue directly.

Will the TODOs be fixed, or migrated as issues, or just remain cuddled in the code for future explorers to marvel at?

dontcallmedom:
FYI: force pushing amended changes means I won't be able to track what's new since my review

src/cli.js
tidoust: I wonder whether that depends on the platform :) In particular, that is how I need to write it in my Windows bash to get Commander to report the string `{ "nightly": { "sourcePath": "compatibility.bs" }  }" that I can then parse into valid JSON. Using single quotes to wrap the parameter does not work on my end.

dontcallmedom: https://stackoverflow.com/a/15262019 suggests that this is indeed a thing in Windows shell. It's not on Unix though :)

.github/ISSUE_TEMPLATE/suggest-spec.yml
```diff
@@ -0,0 +1,40 @@
...
+      value: "https://"
+    validations:
+      required: true
+
+  - type: textarea
+    id: rationale
+    attributes:
+      label: Rationale
+      description: |
+        Please explain how the spec meets the [spec selection criteria](https://github.com/w3c/browser-specs/#spec-selection-criteria), e.g., with a pointer to an intent to implement or to a sister project issue. You may leave this field blank if that seems straightforward.
```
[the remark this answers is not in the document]
tidoust: That's one of the main properties that we may need to capture, but it still only applies to ~6% of entries for the time being, so I propose to add it afterwards. FWIW, the properties that we need to set the most (at least today, these may change over time) and that would be good candidates to collect in a more user-friendly way than through a JSON object are:
- `shortTitle` (49/634)
- `cagegories` (42/634)
- `shortname` (41/634)
- `nightly.sourcePath` (32/634)
- `seriesComposition` (30/634)

src/cli.js
tidoust: I adjusted the text to note that the exact syntax depends on the platform!

tidoust:
I fixed a few hiccups in the jobs YAML files, roughly updated the readme, created @browser-specs-bot, and added the required token that should make jobs update the repo under its name. Let's give this a try ;)
I'll create follow-up issues for possible improvements currently marked as TODOs within the code.

PR #1330, opened 2024-05-19
Handle FPWD for nightly, fix orientation-event series info, log filename responses

tidoust:
The updates fix a couple of test failures that currently prevent builds, and should provide more info about the remaining one (lack of `filename`).


tidoust:
hot fix, merging to get additional logs with next build...

PR #1338, opened 2024-05-28
Move WICG Compression to a WHATWG spec

dontcallmedom:


tidoust (approved):
Looks good for browser-specs. I also created https://github.com/WICG/admin/pull/189 to drop the entry from the WICG biblio so that SpecRef eventually picks up the new URL once it "Compression" has reached the WHATWG biblio.

PR #1341, opened 2024-05-31
Force update of index.json, adjusting discontinued entry

tidoust:
The build code gets lost with the `custom-state-pseudo-class` entry because the URL now redirects to a section of the HTML spec. Goal is to make the entry discontinued. Once that is done, the build code should reuse the info from index.json directly (the discontinued state is a frozen state).

This update forces the standing of the entry to discontinued, reusing previous info for other entry properties. This should fix the issue at hand and allow build to resume.

The other updates are triggered by recent changes to `specs.json` that have not yet been integrated in `index.json` because the build could not succeed. No manual changes were made to these updates.

### DIFF
--- a/index.json
+++ b/index.json
@@ -288,6 +288,46 @@
     }
   },
   {
+    "url": "https://compression.spec.whatwg.org/",
+    "seriesComposition": "full",
+    "shortname": "compression",
+    "series": {
+      "shortname": "compression",
+      "currentSpecification": "compression",
+      "title": "Compression Streams",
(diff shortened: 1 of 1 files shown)


tidoust:
CI tests still fail because the job does not manage to retrieve the SVG2 spec to test pages extraction for some reason. That's totally unrelated. I'll merge as-is. It's likely that the job that looks at creating a release PR will fail as well for the same reason. That should be temporary...

PR #1345, opened 2024-06-03
Make `multipage` flag specifically target release and/or nightly

tidoust:
The job still fails to load the ED of the CSS 2.2 spec to extract the list of pages, even with optimized loading (added in last commit). Something's rotten in the State of Network.

In practice, build does not *need* to load the ED of the CSS 2.2 spec. It just does it because the spec is identified as multipage due to the /TR version. To avoid unnecessary spec loads, this update converts the `multipage` flag from a boolean to a string with 3 possible values: "all", "release", "nightly". The extraction of the list of pages is now only done on the version targeted by the `multipage` property.

The update also serializes the extraction of pages to avoid sending too many requests at once. There are only a handful of specs for which we need to look for pages, serialization should not make build much longer than it already is.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,9 +120,12 @@ automatically is not the expected one.
 - `forceCurrent`: a boolean flag to tell the code that the spec should be seen
 as the current spec in the series. The property must only be set when value is
 `true`.
-- `multipage`: a boolean flag to identify the spec as a multipage spec. This
+- `multipage`: a flag to identify the spec as a multipage spec. This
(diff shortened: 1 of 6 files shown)


dontcallmedom (approved):
note that at least one of the recent failures was for the SVG2 spec rather than CSS2

tidoust:
> note that at least one of the recent failures was for the SVG2 spec rather than CSS2

Yes, and the situation is worse in Webref, we just don't see it ;)
I created https://github.com/w3c/webref/issues/1244 to explore and track the issue.

PR #1353, opened 2024-06-05
Always reuse previously determined filenames

tidoust:
This update restricts the actual determination of filenames to new specs, to avoid network requests. When network requests need to be sent, they are now sent sequentially, with a 1s delay between requests. There should be a handful of specs for which network requests are needed at most.

### DIFF
--- a/src/build-index.js
+++ b/src/build-index.js
@@ -341,37 +341,29 @@ async function runPages(index) {
 
 
 async function runFilename(index, { previousIndex, log }) {
+  for (const spec of index) {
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):
with that new approach, would it be worth reverting the w3cTr shortcut in determine-filename.js?

tidoust:
> with that new approach, would it be worth reverting the w3cTr shortcut in determine-filename.js?

I kept it in the interest of avoiding all possible network requests, even when talking about a single request to a /TR URL. But it's more me being pessimistic here. That should be pretty useless in practice indeed.

PR #1354, opened 2024-06-06
Limit test on builds to data

dontcallmedom:
The build is failing because it is running the full test suite which makes HTTP requests that fail; while that is probably worth addressing on its own, it is not necessary to test the code when it is only the data we're updating, which this PR enables.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        npm run test-data
+        npm run test-index
 
     - name: Bump minor version of packages if needed
       run: node src/bump-packages-minor.js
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        npm run test
+        npm run test-data
```
tidoust, line 37:
```suggestion
        npm run test-index
```

package.json
```diff
@@ -14,7 +14,8 @@
     "build": "node src/build-index.js",
     "lint": "node src/lint.js",
     "lint-fix": "node src/lint.js --fix",
-    "test": "mocha"
+    "test": "mocha",
+    "test-data": "mocha test/index.js"
```
tidoust, line 18:
```suggestion
    "test-index": "mocha test/index.js"

```

tidoust (approved):
Just suggest to rename to `test-index` to be clearer about what is being tested

PR #1356, opened 2024-06-06
Add throttling logic per origin

tidoust:
[Creating as draft as I'd be more confident if I were able to run a successful build locally]

This adds a generic throttled queue per origin class that replaces the previous `throttle` function to add serialization (and sleep interval) per origin.

The code is inspired by:
https://github.com/w3c/cg-monitor/blob/5ceea24a1eaa7c1736d38909bcbb47df4ae297a3/lib/caching-queued-fetch.js#L68

There may be a way to rewrite the class in a function wrapper as the previous throttle module was doing. I just grew tired of abstractions.

(Note plan is to reuse the same class in Reffy, see https://github.com/w3c/reffy/pull/1587)


src/fetch-info.js
```diff
@@ -206,7 +207,7 @@ async function fetchInfoFromSpecref(specs, options) {
   // API does not return the "source" field, so we need to retrieve the list
   // ourselves from Specref's GitHub repository.
   const specrefBrowserspecsUrl = "https://raw.githubusercontent.com/tobie/specref/main/refs/browser-specs.json";
-  const browserSpecsResponse = await throttledFetch(specrefBrowserspecsUrl, options);
+  const browserSpecsResponse = await fetch(specrefBrowserspecsUrl, options);
```
tidoust, line 210: Note throttling is useless there, since we're sending only one such request.

src/fetch-info.js
```diff
@@ -303,7 +304,7 @@ async function fetchInfoFromIETF(specs, options) {
   }
 
   async function fetchRFCName(docUrl) {
-    const res = await fetch(docUrl, options);
+    const res = await fetchQueue.runThrottled(fetch, docUrl, options);
```
tidoust, line 307: ... while here, throttling is more useful.

src/fetch-info.js
```diff
@@ -361,6 +362,7 @@ async function fetchInfoFromIETF(specs, options) {
     return paths.filter(p => p.path.match(/^specs\/rfc\d+\.html$/))
       .map(p => p.path.match(/(rfc\d+)\.html$/)[1]);
   }
+  const httpwgRFCs = await getHttpwgRFCs();
```
tidoust, line 365: I'm rolling back on a change I introduced last time to avoid fetching this when that was not needed. I was drunk, the change only resulted in the request being sent for each and every IETF spec that was processed...

tidoust:
I'm marking the PR as ready for review, but then I still haven't been able to test it for real with the borked CSS drafts specs server, and so do not know the impact on the build duration for now.

src/throttled-queue.js
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * Additionally, the function forces a 2 second sleep after processing to
+   * keep a low network profile.
+   */
+  async runThrottledPerOrigin(url, processFunction, ...params) {
+    const origin = getOrigin(url);
+    if (!this.originQueue[origin]) {
+      this.originQueue[origin] = Promise.resolve(true);
+    }
+    return new Promise((resolve, reject) => {
```
dontcallmedom, line 109: Using `Promise.withResolvers` would probably help with readability

dontcallmedom:
While we're looking at the cg-monitor, another idea from there maybe worth exploring is to build an HTTP cache and reuse it from one job to the next with GitHub caches.

src/throttled-queue.js
tidoust: Never heard of that ;)

But this ([baseline 2024](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers)) function was apparently added in Node.js 22.0.0, not in Node.js LTS for now!

tidoust:
> [Creating as draft as I'd be more confident if I were able to run a successful build locally]

Managed to run a build locally this morning. The fetch-info step took about ~8mn (compared to ~4mn previously). That's not surprising for >200 specs, most of them being CSS drafts or `github.io` specs. If we need to speed things up a bit, we may want to treat `xxx.github.io` URLs as different origins, but I propose to err on the conservative side for now.

Anyway, that takes longer but that worked. Meanwhile, the job failed, this time with a 429 on a CSS stylesheet served from w3.org. I think the only reason why we keep loading CSS stylesheets is that ReSpec loads a few styles dynamically and may not manage to build the spec if styles cannot be loaded. That may be worth checking again though.

dontcallmedom:
how about we intercept css requests but instead of failing them, we return empty content?

dontcallmedom:
that being said, the last job I see seemed to have 429-failed on the fxtf server, not on the CSS stylesheet

tidoust:
> that being said, the last job I see seemed to have 429-failed on the fxtf server, not on the CSS stylesheet

Ah right, the protocol error on the CSS stylesheet is probably a consequence of the build stopping due to the 429, not the cause.

plinss:
The rate limiting on the csswg.org server is fairly recent (note that the same machine serves fxtf.org and css-houdini.org).

It's currently set to 1 request per second but allows bursts to load associated resources (varies based on path, e.g. more for images).

The rate limiting was implemented to defend against several crawlers that were severely impacting the server, but it can be adjusted if needed.

Also note that the server is also now running CrowdSec and repeat rate limit violations will result in temporary bans at the firewall (4 hours at first, increasing an additional 4 hours per offense).

tidoust:
Thanks for clarifying the rate limit rules that were put in place, @plinss!

I guess browser-specs and Reffy were among the impacting crawlers ;) We have now tamed our crawling beasts, and the rules we adopted should work well with the limits you mention:

- The crawler only sends one request at a time to a given origin
- The crawler sleeps 2s in between requests sent to the csswg.org server
- The crawler avoids loading associated resources (CSS stylesheets, images, some scripts that we know we do not need)
- `csswg.org`, `fxtf.org` and `css-houdini.org` are considered to be the same origin

Crawls have resumed without errors since yesterday, so all looks good!

PR #1359, opened 2024-06-07
Use fallback data on HTTP error

dontcallmedom:


src/fetch-info.js
```diff
@@ -57,7 +57,7 @@ const specrefStatusMapping = {
 
 const fetchQueue = new ThrottledQueue(2);
 
-async function useLastInfoForDiscontinuedSpecs(specs) {
+function useLastInfoForDiscontinuedSpecs(specs) {
```
tidoust, line 60: That's fine but I thought I'd note that I think about it differently: this is one of the step functions, some of the step functions are asynchronous, I'd like all of them to be asynchronous for consistency, as such I want them to return a `Promise` which the `async` keyword guarantees.

src/fetch-info.js
```diff
@@ -68,6 +68,19 @@ async function useLastInfoForDiscontinuedSpecs(specs) {
   return results;
 }
 
+function catchAndFallbackOnExistingData(fn) {
+  return function(spec) {
+    return fn(spec).catch(err => {
+      if (spec.__last) {
+	// TODO: log crawl error?
```
tidoust, line 75: We really need to log an error somewhere, be it to the console. Otherwise we won't even be able to tell that we're reusing the last information. That's a very good way to hide the problem, but means we could get stuck with hidden build failures for months.

tidoust (changes requested):

src/fetch-info.js
dontcallmedom, line 60:
```suggestion
async function useLastInfoForDiscontinuedSpecs(specs) {
```
fair :)

src/fetch-info.js
dontcallmedom, line 75:
```suggestion
	// TODO: log crawl error more visibly?
	console.error(`Failed to fetch info on ${spec.url} (${err}), reusing existing data`);
```
This adds console logging; I do agree we need something more robust to detect these situations, but maybe that's a sufficient starting point.

PR #1360, opened 2024-06-07
Refresh throttling logic (based on code used in Reffy)

tidoust:
This imports the adjustments made to the throttling logic in Reffy. In the end, it seems useful to adopt the same strategy regarding sleeping between requests: 2s for CSS server, 1s for www.w3.org, and 100ms for other origins.

This raises the question of whether that logic should rather be integrated into throttled-queue.js if we use the same logic everywhere. Oh well ;)

Refresh also skips CSS resources as done in Reffy.

### DIFF
--- a/src/fetch-info.js
+++ b/src/fetch-info.js
@@ -55,7 +55,7 @@ const specrefStatusMapping = {
   "cg-draft": "Draft Community Group Report"
 };
 
-const fetchQueue = new ThrottledQueue(2);
+const fetchQueue = new ThrottledQueue({ maxParallel: 2 });
 
 async function useLastInfoForDiscontinuedSpecs(specs) {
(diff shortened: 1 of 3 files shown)


dontcallmedom (approved):
sounds like we need a crawl-web-spec library >:)

tidoust:
> sounds like we need a crawl-web-spec library >:)

Hélas! Tracked in https://github.com/w3c/reffy/issues/1589

plinss:
See: https://github.com/w3c/browser-specs/pull/1356#issuecomment-2155481520

Note that the rate limits for csswg.org also apply to fxtf.org and css-houdini.org. The base rate limit is 1 request per second, but images and stylesheets also count (bursts are allowed).

PR #1389, opened 2024-06-26
Make css-overflow-5 a delta spec, add related test

tidoust:
The css-overflow-5 spec was added last week as a full spec, but css-overflow-4 is a delta spec. On top of being incorrect because level 5 is also a delta spec, the succession full > delta > full confuses Reffy, who considers that level 5 is the "latest version" it should report extracts for, instead of level 3: it ends up with creating a `css/css-overflow.json` extract for level 5, `css/css-overflow-4.json` for level 4, and no extract for level 3.

This update makes css-overflow-5 a delta spec and adds a test to guarantee that a delta spec can never be followed by a full spec in its series.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -113,7 +113,7 @@
     "seriesComposition": "delta",
     "shortTitle": "CSS Multicol 2"
   },
-  "https://drafts.csswg.org/css-overflow-5/",
+  "https://drafts.csswg.org/css-overflow-5/ delta",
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):
thanks, and sorry I missed the `delta` annotation when reviewing it last week

tidoust:
(I note CI tests fail precisely because the new test detects the problem in the current list)

PR #1402, opened 2024-07-10
Update repository for core-aam

tidoust:
The source of the spec was moved to the main ARIA GitHub repository. The Editor's Draft is still being served from the previous repository though, meaning the URL of the Editor's Draft remains unchanged.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -882,7 +882,7 @@
     "url": "https://www.w3.org/TR/core-aam-1.2/",
     "nightly": {
       "repository": "https://github.com/w3c/aria",
-      "sourcePath": "accname/index.html"
+      "sourcePath": "core-aam/index.html"
     }
   },
   "https://www.w3.org/TR/credential-management-1/",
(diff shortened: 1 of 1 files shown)


specs.json
```diff
@@ -878,7 +878,13 @@
   },
   "https://www.w3.org/TR/compute-pressure/",
   "https://www.w3.org/TR/contact-picker/",
-  "https://www.w3.org/TR/core-aam-1.2/",
+  {
+    "url": "https://www.w3.org/TR/core-aam-1.2/",
+    "nightly": {
+      "repository": "https://github.com/w3c/aria",
+      "sourcePath": "accname/index.html"
```
dontcallmedom, line 885:
```suggestion
      "sourcePath": "core-aam/index.html"
```
shouldn't it be "core-aam/index.html" instead?

tidoust: Ahem. Let's say that was a PR review test ;)

dontcallmedom: Yeah, I passed!

PR #1421, opened 2024-07-26
Fix parsing of github.com pages in find-specs script

tidoust:
The script was expecting tables to follow their headings at the same level in the HTML, but GitHub now wraps tables in `<markdown-accessibility-table>` web components, and headings in a `<div>`.


dontcallmedom (approved):
good find!

PR #1424, opened 2024-07-26
Add `encrypted-media` as former name for `encrypted-media-2`

tidoust:
Missed during previous update.

Arguably, the continuity exists through the series shortname, but the guarantee that we make is that any shortname that used to appear will either appear as a shortname or in a `formerNames` property.


dontcallmedom (approved):

tidoust:
Note CI tests continue to fail because the PR does not update `index.json`. This will pass once a new build gets done.

PR #1426, opened 2024-07-29
ESM all the things!

tidoust:
Note that, while Mocha supports ESM modules when used as CLI, it does not support them out of the box when using Mocha programmatically. The code forces the loading of test modules as ESM through `loadFilesAsync()` before it calls `run()`, as suggested in the doc:
https://mochajs.org/api/mocha#loadFilesAsync

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,8 @@
 "use strict";
 
-const specs = require("./index.json");
+import specs from "./index.json" with { type: "json" };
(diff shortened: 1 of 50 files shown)


tidoust:
The updates are similar to those done in Reffy and Webref. I'm taking the liberty to merge as-is, and will iterate on the code as needed.

PR #1464, opened 2024-08-22
Add Isolated Contexts spec to specs.json

robbiemc:


dontcallmedom:
hi @robbiemc thanks for the PR - the preferred way to add specs to browser-specs is now to use the issue template, since that allows to detect more issues; I've done so in #1466 and generated #1467 out of it - note in particular that I'm suggesting to *not* add (currently) the spec as a spec expected to be implemented in browsers since its deployment seems so far to be limited to ChromeOS.

I've also filed https://github.com/WICG/isolated-web-apps/pull/49 to help reducing the risks that the monkey patches in the spec interfere with the data extracted from the current specs.

robbiemc:
Thanks for creating the issue.

Isolated Contexts are currently implemented cross-platform in chromium, and can be enabled with a flag on all desktop platforms, but are only launching on ChromeOS for the short to medium term. We plan on eventually launching to on other platforms in the future. It is intended to be a browser feature, just not enabled on all platforms for now.

I'm not opposed to removing the "browser" category if that generally implies enablement cross-platform, I just want to give more context.

tidoust:
The spec was added to the list (will appear in browser-specs later today). The `-browser` setting can be dropped when the spec makes progress towards other platforms.

PR #1486, opened 2024-09-03
update FedCM to W3C FPWD

elf-pavlik:
FedCM spec advanced from CG to WG, which published a First Public Working Draft.

related https://github.com/mdn/browser-compat-data/pull/24289

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -138,7 +138,7 @@
   },
   "https://drafts.fxtf.org/filter-effects-2/ delta",
   "https://encoding.spec.whatwg.org/",
-  "https://fedidcg.github.io/FedCM/",
+  "https://www.w3.org/TR/fedcm/",
   "https://fetch.spec.whatwg.org/",
   {
     "url": "https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html",


elf-pavlik:
Searching for https://github.com/search?q=org%3Amdn+fedidcg.github.io&type=code

also shows two occurrences in mdn/rumba

tidoust:
Thanks for the report! A job usually alerts us when a first public working draft is published, but the code did not know anything about the `w3c-fedid` organization on GitHub and missed that one.

On top of the linting error (entries are sorted by URL), the update also needs to record the shortname change, from "fedCM" to "fedcm-1", using a `formerNames` property. I'll propose the right change in a separate pull request.

tidoust:
Done in #1487.

> also shows two occurrences in mdn/rumba

I suppose the new URL will propagate itself to mdn/rumba automagically.

Side note: adding a redirect to the new URL of the Editor's Draft would avoid the 404. For example, the WICG usually does that for specs that transition to other groups, see: https://github.com/WICG/wicg.github.io. The redirects are not perfect (at the HTML level, not at the HTTP level), but that avoids breaking links.

specs.json
```diff
@@ -138,7 +138,7 @@
   },
   "https://drafts.fxtf.org/filter-effects-2/ delta",
   "https://encoding.spec.whatwg.org/",
-  "https://fedidcg.github.io/FedCM/",
+  "https://www.w3.org/TR/fedcm/",
   "https://fetch.spec.whatwg.org/",
   {
     "url": "https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html",
```
Kazh77net, line 144: /

Kazh77net (approved):

PR #1571, opened 2024-11-20
Add RFC 6455: The WebSocket Protocol

bsmth:
This PR adds [RFC 6455](https://www.rfc-editor.org/rfc/rfc6455), including a `shortTitle` for convenience.

This is required for adding MDN browser compat data for the `Sec-WebSocket-*` HTTP headers.

__Related issues and pull requests:__

- [ ] https://github.com/mdn/browser-compat-data/pull/25115

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -820,6 +820,10 @@
     "url": "https://www.rfc-editor.org/rfc/rfc6266",
     "shortTitle": "Content-Disposition in HTTP"
   },
+  {
+    "url": "https://www.rfc-editor.org/rfc/rfc6455",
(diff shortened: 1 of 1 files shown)


bsmth:
> That workflow allows us to see and check the resulting entry before we add the spec to the list, and to create a PR automatically afterwards.

Ah, I understand, thanks for clarifying.

>I went through it, see https://github.com/w3c/browser-specs/issues/1574, leading to https://github.com/w3c/browser-specs/pull/1575. Spec will start appearing in browser-specs next time a build succeeds, hopefully in a few hours.

Super, thank you very much!

PR #1658, opened 2025-01-21
Don't fetch info from Specref for CSS and www.w3.org specs.

tidoust:
Browser-specs still fetched info from Specref for a few W3C specs, notably the `css-backgrounds-4` draft and a few policy documents. Info is now fetched directly from the spec for these specs. This will make it possible to properly retire `csswg.json` in Specref.

The GIF spec, published under `https://www.w3.org` is treated as an exception to the rule, because its URL ends with `.txt`, signaling a spec published in plain text.

Note: Browser-specs still leverages Specref, mainly for WHATWG specs, a few TC39 and ISO specs, and a couple of CG specs that use their own GitHub org.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -879,6 +879,7 @@
     "url": "https://www.w3.org/Graphics/GIF/spec-gif89a.txt",
     "shortname": "GIF",
     "shortTitle": "GIF",
+    "title": "Graphics Interchange Format",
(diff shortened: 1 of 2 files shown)


tidoust:
Note CI tests fail because `iso.org` is currently unstable. Unrelated to the update!

dontcallmedom (approved):

PR #1664, opened 2025-01-23
Update URL of Web Speech API

tidoust:
The update preserves the former shortname `speech-api` for now, matching the
`shortname` metadata currently present in the spec.

Fixes https://github.com/w3c/browser-specs/issues/1639


tidoust:
Screwed up the history of commits locally but the update's good ;)

dontcallmedom (approved):

PR #1666, opened 2025-01-24
Retrieve WHATWG spec titles from WHATWG database

tidoust:
Build code used to rely on Specref to get the title of WHATWG specifications. This update makes it fetch info for WHATWG specs from the WHATWG database directly. To save one request, the code leverages the workstreams database, also used by fetch-groups, instead of the biblio file.

On top of adding a new `whatwg` value to the `"source"` field, this update will also fix the titles of the WHATWG specs: they end with "Standard" in Specref but, while that matches the `<title>` tag, the actual spec title in the `<h1>` and the title in the WHATWG database don't end with "Standard". *#docallmeDOM*

The update turns the `fetchJSON` function into a utility function. This is going to save a few requests (not that many!) that are common between fetch-info and fetch-groups. Specific functions in fetch-info were also adjusted not to do anything when there are no specs of interest in the list (this speeds up tests a bit, but has no impact on a full build since, by definition, there are specs of interest in the full list...)

### DIFF
--- a/src/fetch-json.js
+++ b/src/fetch-json.js
@@ -1,0 +1,33 @@
+import ThrottledQueue from "./throttled-queue.js";
+
+// Make sure we remain "friendly" with servers
+const fetchQueue = new ThrottledQueue({ maxParallel: 2 });
+
+// Maintain a cache of fetched JSON resources in memory to avoid sending the
+// same fetch request again and again
+const cache = {};
+
+/**
+ * Fetch a JSON URL
+ */
+export default async function (url, options) {
+  if (cache[url]) {
+    return JSON.parse(JSON.stringify(cache[url]));
+  }
+  const res = await fetchQueue.runThrottled(fetch, url, options);
+  if (res.status === 404) {
+    return null;
+  }
+  if (res.status !== 200) {
+    throw new Error(`Server returned an error for ${url}, status code is ${res.status}`);
+  }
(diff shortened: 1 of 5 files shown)


src/fetch-json.js
```diff
@@ -0,0 +1,33 @@
...
+// Maintain a cache of fetched JSON resources in memory to avoid sending the
+// same fetch request again and again
+const cache = {};
+
+/**
+ * Fetch a JSON URL
+ */
+export default async function (url, options) {
+  if (cache[url]) {
+    return JSON.parse(JSON.stringify(cache[url]));
```
dontcallmedom, line 15:
```suggestion
    return structuredClone(cache[url]);
```

src/fetch-json.js
```diff
@@ -0,0 +1,33 @@
...
+    return null;
+  }
+  if (res.status !== 200) {
+    throw new Error(`Server returned an error for ${url}, status code is ${res.status}`);
+  }
+
+  try {
+    const body = await res.json();
+    cache[url] = body;
+    return JSON.parse(JSON.stringify(body));
```
dontcallmedom, line 28:
```suggestion
    return structuredClone(body);
```

dontcallmedom (approved):
minor suggested improvement; very nice code clean up overall!

PR #1668, opened 2025-01-24
Stop fetching spec info from Specref

tidoust:
This completely drops the logic that used to fetch spec info from Specref. Motivation for the change is to avoid having a circular dependency between browser-specs and Specref that makes it cumbersome to make some updates.

In practice, there remained only 6 specs for which browser-specs was leveraging Specref: 3 from ISO, 2 from TC39 and the Web Bluetooth API.

One consequence is that the title of the TC39 specs will change because the actual specs include a registered mark and the year, which are now captured, leading to "ECMAScript® 2025 Language Specification". Maybe not ideal, but then that's per spec...

README and comments updated in the code to remove mentions of Specref.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,11 @@ The list is used in a variety of ways, which include:
   by tools such as [ReSpec](https://respec.org/docs/) and
   [Bikeshed](https://speced.github.io/bikeshed/) to create terminology
   and reference links between Web specifications.
+* [BCD](https://github.com/mdn/browser-compat-data) and
+  [web-features](https://github.com/web-platform-dx/web-features) to validate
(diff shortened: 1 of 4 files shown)


tidoust:
Note: CI tests fail because there remain specs in `index.json` with a `specref` source whereas I removed that value from the schema. Tests will pass once this is merged and a new build succeeds.

tidoust:
> hmm... I'm a bit confused - how are going to fetch info for ISO specs moving forward?

From the ISO pages. We already have that logic in place and we just have 3 ISO specs in the list for now.

dontcallmedom (approved):

PR #1732, opened 2025-03-13
Drop release URL for CSS2 (URL was fixed in the W3C API)

tidoust:
The W3C API now returns the /TR/CSS2/ URL for CSS2, the URL no longer needs to be specified in `specs.json`.

Note: this is the expected follow-up to #1726


tidoust:
CI tests fail due to CSS Houdini server being temporarily down. That's unrelated.

dontcallmedom (approved):

PR #1806, opened 2025-04-23
Update ECMA-426 source reference

Bashamega:
The TC39 committee has updated the source of ECMA-426 to source-map-rev3.md.


tidoust (approved):
Thanks!

PR #1807, opened 2025-04-23
Update webappsec-passkey-endpoints source reference

Bashamega:


Bashamega:
Thank you very much @tidoust, for your review. I have updated it accordingly

specs.json
```diff
@@ -545,13 +545,8 @@
...
   },
-  "https://w3c.github.io/webappsec-dbsc/",
-  {
-    "nightly": {
-      "sourcePath": "passkey-endpoints.bs"
-    },
-    "url": "https://w3c.github.io/webappsec-passkey-endpoints/passkey-endpoints.html"
-  },
+  "https://w3c.github.io/webappsec-passkey-endpoints/",
+  "https://w3c.github.io/webappsec-passkey-endpoints/passkey-endpoints.html",
```
tidoust, line 549: There should be only one!

```suggestion
```

tidoust (changes requested):

specs.json
```diff
@@ -545,13 +545,7 @@
     "url": "https://w3c.github.io/web-share-target/",
     "standing": "good"
   },
-  "https://w3c.github.io/webappsec-dbsc/",
```
tidoust, line 548: Almost there ;) I hadn't realized that you had deleted the `webappsec-dbsc` entry. That's a different spec, it needs to stay!

Bashamega: Done

tidoust (approved):
Thanks!

PR #1831, opened 2025-04-30
Leverage validate-repos and throttle requests on w3c.json files

tidoust:
The validate-repos project collects all known `w3c.json` file, and contains
the information we need to map repo names to groups for a good chunk of specs.

This still does not work well for /TR specs because we guess the name of the
repository from the spec's shortname, but that's still better than what we used
to have!

Also, the `fetchJSON` function was only applying loose throttling, but
https://raw.githubusercontent.com/ is starting to return 429 responses after a
while. Documentation does not say much, but it would seem the limit is around
~5000 requests per hour. This update makes the function sleep 1000ms in
between requests to a given origin.

### DIFF
--- a/src/fetch-groups.js
+++ b/src/fetch-groups.js
@@ -12,6 +12,13 @@ import fetchJSON from "./fetch-json.js";
 
 
 /**
+ * We will very likely need to use group information from the validate-repos
+ * project which compiles w3c.json files across repositories.
+ */
+let w3cGroups = null;
+
+
+/**
  * Retrieve the information about the exact organization and the group that
  * develops an ISO specification from the description page on the ISO web site.
  *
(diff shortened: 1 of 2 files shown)


src/fetch-groups.js
```diff
@@ -195,9 +202,29 @@ export default async function (specs, options) {
     spec.organization = spec.organization ?? "W3C";
 
     if (!spec.groups) {
-      let groups = null;
+      // Get group info from validate-repos if possible to avoid having to
+      // send individual network requests for each spec
+      // Note: this will not yield anything for many /TR specs because we
+      // guess the name of the repo from the shortname.
```
dontcallmedom, line 208: would https://github.com/w3c/spec-dashboard/blob/gh-pages/repo-map.json help?

dontcallmedom (approved):

src/fetch-groups.js
tidoust: Oh yes, I had forgotten about this one (which find-specs also makes use of).

tidoust: Tracked in #1832

PR #1851, opened 2025-05-12
Add new RDF/SPARQL WG to list of "non-browser" groups

tidoust:
The group used to be known as the "RDF-star" working group. New name needed to be added to the list of non-browser groups to avoid categorizing the group's deliverables as "browser" specs.

Note: the list still contains the previous name in case some data remains that still references that name.

Via https://github.com/w3c/browser-specs/pull/1828#pullrequestreview-2820568835


tidoust:
Taking the liberty to merge to unlock publication of the package.

PR #1900, opened 2025-05-26
Drop dependency on Mocha in CLI

### LINKED ISSUE #1899: Fix browser-specs CLI (still depends on Mocha)
The CLI code directly calls on Mocha to run tests against the would-be index file. Mocha was dropped as a dependency but this code was not updated. The bot currently fails to check spec proposals as a result.

tidoust:
This replaces the CLI logic that was based on Mocha for running tests against the index that a proposed change would create, with logic based on Node's test module.

Note: I added a simple test on the CLI to make sure that it's at least possible to run it, so that we don't miss that next time. More tests would of course be better...

Fixes #1899.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -3,12 +3,13 @@ import { Command } from 'commander';
 import path from 'node:path';
 import { fileURLToPath } from "node:url";
 import fs from 'node:fs/promises';
-import Mocha from 'mocha';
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):
I see that I had left the core of the migration as an exercise to the reader - sorry :)

PR #2035, opened 2025-07-22
Use ISO Open Data datasets to process ISO specs

tidoust:
Logic to process specs from ISO was to "browse" the ISO web site to extract the spec title and group name. That approach fails more and more, essentially because bots are not welcome on the web site.

This update drops custom logic to browse the ISO web site and leverages machine-readable datasets provided through the ISO Open data project instead: https://www.iso.org/open-data.html

Now, the thing is, browser-specs only contains 3 specs from ISO for now, and their information is unlikely ever going to change. And the datasets are huge (~70MB for the main catalog). To avoid wasting resources for nothing, a new build command gets introduced:

```
npm run build-skip-iso
```

The command skips fetching the ISO catalog and reuses previous info about the ISO specs.

That new command is used for all automated builds from Monday to Friday. On Saturday and Sunday, a full build gets made, only once per day (running multiple builds during the week-end is not very useful in any case as we're unlikely ever going to act on them).

Under the hoods, the new command uses a `--skip-fetch` CLI option, which could perhaps be extended afterwards to skip additional network fetches.

The `source` of ISO entries changes from `spec` to `iso`.

Note: A full build will be needed when we add a new ISO spec to the list.

### DIFF
--- a/src/build-index.js
+++ b/src/build-index.js
@@ -27,6 +27,7 @@ import computeStanding from "./compute-standing.js";
 import determineFilename from "./determine-filename.js";
 import determineTestPath from "./determine-testpath.js";
 import extractPages from "./extract-pages.js";
+import fetchISOSpecs from "./fetch-iso-info.js";
 import fetchInfo from "./fetch-info.js";
 import fetchGroups from "./fetch-groups.js";
 import loadJSON from "./load-json.js";
@@ -54,6 +55,11 @@ const steps = [
     shortname: "previous",
     title: "Retrieve last published info",
     run: runFetchLastPublished
+  },
+  {
+    shortname: "iso",
+    title: "Process ISO specs",
+    run: (index, options) => fetchISOSpecs(index, options)
   },
   {
     shortname: "groups",
@@ -249,7 +255,7 @@ async function runInfo(specs) {
 
     // Set the series title based on the info returned by the W3C API if
     // we have it, or compute the series title ourselves
-    const seriesInfo = specInfo.__series[spec.series.shortname];
+    const seriesInfo = specInfo.__series?.[spec.series.shortname];
     if (seriesInfo?.title && !res.series.title) {
       res.series.title = seriesInfo.title;
     }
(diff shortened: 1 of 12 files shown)


src/build-index.js
```diff
@@ -396,35 +407,43 @@ async function generateIndex(specs, { step = "all", previousIndex = null, log =
...
 
 /*******************************************************************************
-Export functions for use as module
+Export main function for use as module
 *******************************************************************************/
-export {
-  generateIndex,
-  generateIndexFile
-};
+export { generateIndex };
```
tidoust, line 428: Note: there was no reason to export `generateIndexFile`. Nothing uses it.

src/build-index.js
```diff
@@ -249,7 +255,7 @@ async function runInfo(specs) {
 
     // Set the series title based on the info returned by the W3C API if
     // we have it, or compute the series title ourselves
-    const seriesInfo = specInfo.__series[spec.series.shortname];
+    const seriesInfo = specInfo.__series?.[spec.series.shortname];
```
tidoust, line 258: Needed when running a "diff build" that only has ISO specs as `__series` is only set when there are W3C specs.

src/fetch-info.js
```diff
@@ -37,18 +37,21 @@
 
 import puppeteer from "puppeteer";
 import loadSpec from "./load-spec.js";
-import computeShortname from "./compute-shortname.js";
```
tidoust, line 40: Note: this was a useless dependency, the function was never called.

src/fetch-iso-info.js
```diff
@@ -0,0 +1,102 @@
...
+
+  const isoRe = /\.iso\.org\//;
+  const isoSpecs = specs.filter(spec => spec.url.match(isoRe));
+  if (isoSpecs.length === 0) {
+    return specs;
+  }
+
+  if (!['all', 'iso'].includes(options?.skipFetch)) {
+    // Fetch the list of technical committees
+    const tcUrl = 'https://isopublicstorageprod.blob.core.windows.net/opendata/_latest/iso_technical_committees/json/iso_technical_committees.jsonl';
```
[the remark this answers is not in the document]
dontcallmedom: makes sense

PR #2132, opened 2025-09-04
Update wot-binding-registry url in specs.json

vmpstr:
wot-bindings-registry (plural bindings) doesn't seem to exist, which may have been a typo

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -622,8 +622,8 @@
   "https://w3c.github.io/webrtc-ice/",
   "https://w3c.github.io/wot-binding-registry/",
   {
-    "shortTitle": "WoT Bindings Registry",
-    "url": "https://w3c.github.io/wot-bindings-registry/"
+    "shortTitle": "WoT Binding Registry",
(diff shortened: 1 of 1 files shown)


specs.json
```diff
@@ -621,10 +621,6 @@
   },
   "https://w3c.github.io/webrtc-ice/",
   "https://w3c.github.io/wot-binding-registry/",
```
tidoust, line 623:
```suggestion
  {
    "url": "https://w3c.github.io/wot-binding-registry/",
    "shortTitle": "WoT Bindings Registry",
    "formerNames": [
      "wot-bindings-registry"
    ]
  },
```

tidoust (commented):
Ouch, the underlying GitHub repository was renamed and I missed that we already had a previous entry when I reviewed the new name. The former name `wot-bindings-registry` needs to be recorded in a `formerNames` property.

I also missed the fact that the `shortTitle` needs to be specified because the code generates "WoT" otherwise, which is too generic.

Thanks for the report!

PR #2156, opened 2025-09-22
Add new version of vocab-ssn, rename old one

tidoust:
The `vocab-ssn` shortname now is an alias to `vocab-ssn-2017`, and series has a new `vocab-ssn-2023` working draft.

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -1881,7 +1881,17 @@
     "nightly": {
       "sourcePath": "ssn/index.html"
     },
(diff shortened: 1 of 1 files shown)


tidoust:
FWIW, I find the diff a tad confusing: the update does not touch `vocab-ssn-ext`, but linting means it gets moved below ` vocab-ssn-2017` and `vocab-ssn-2023`.

PR #2319, opened 2026-01-26
Discontinue TC39 proposals integrated in ECMAScript

tidoust:
This flags TC39 proposals that have been included in the latest draft of the ECMAScript specification in the past 6 months or so, based on: https://github.com/tc39/proposals/blob/main/finished-proposals.md

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -368,7 +368,13 @@
       "ecmascript"
     ]
   },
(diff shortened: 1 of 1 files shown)


dontcallmedom:
thanks for the notice - I have checked and json-parse-with-source is the only one that hadn't gotten merged in the main spec yet, we'll revert that one

PR #2323, opened 2026-01-28
Revert marking JSON Parse with Source as discontinued

dontcallmedom:
Not yet merged in main spec, see also https://github.com/w3c/browser-specs/pull/2319#issuecomment-3807920724

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -539,13 +539,7 @@
       "ecmascript"
     ]
   },
-  {
-    "url": "https://tc39.es/proposal-json-parse-with-source/",
-    "standing": "discontinued",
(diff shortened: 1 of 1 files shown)


caugner:
Thank you!

Would you consider releasing this as web-specs 3.77.1 before the end of the week? No worries if not, then we would address this in BCD.

tidoust:
Now done and released in the latest version of web-specs. I see our misadventure prompted TC39 to update the description in the [finished proposals](https://github.com/tc39/proposals/blob/main/finished-proposals.md) document. I'll check things more thoroughly next time instead of relying on this list alone.

PR #2353, opened 2026-02-15
TC39 Proposal: Add legacy RegExp features proposal URL to specs.json

hamishwillee:
The https://github.com/tc39/proposal-regexp-legacy-features proposal reached stage 3. I need to add it as a spec for Browser compatibility data https://github.com/mdn/browser-compat-data/pull/28999

Hope this is correct approach!

### DIFF
--- a/specs.json
+++ b/specs.json
@@ -172,6 +172,10 @@
       "sourcePath": "source"
     }
   },
+  {
+    "url": "https://github.com/tc39/proposal-regexp-legacy-features",
+    "shortTitle": "Legacy RegExp features in JavaScript"
+  },
   "https://immersive-web.github.io/anchors/",
   "https://immersive-web.github.io/body-tracking/",
   "https://immersive-web.github.io/model-element/",


tidoust:
> The https://github.com/tc39/proposal-regexp-legacy-features proposal reached stage 3. I need to add it as a spec for Browser compatibility data [mdn/browser-compat-data#28999](https://github.com/mdn/browser-compat-data/pull/28999)

So far, browser-specs requires an actual published spec, we cannot just add "specs" that are GitHub repositories, explainers, pull requests, or the like. It may be that browser-specs needs to become more flexible in that regard. We've managed to survive without it until now...

For what it's worth, when a spec exists, [TC39 proposals](https://github.com/tc39/proposals/) get automatically detected and proposed for addition.

> 
> Hope this is correct approach!

Not exactly but not a big deal either :) We use a semi-automated mechanism to assess proposals for addition, and always start from an issue that follows the [New spec issue template](https://github.com/w3c/browser-specs/issues/new/choose). See [How to add a spec](https://github.com/w3c/browser-specs/blob/main/CONTRIBUTING.md#how-to-add-a-spec).

I'm closing this pull request and created #2358 to track the addition. I don't have an immediate solution to propose to add the spec though.

hamishwillee:
Thanks so much for the explanation @tidoust. Will add it as non-standard for now.

Sorry about the format - I should have know this but I do it so rarely. Have sent myself (another reminder) with links to that info.

PR #2459, opened 2026-04-30
Add `specs.json` mechanism to drop computation of repository

tidoust:
This reverts #2457 because there are /TR specs that use auto-publish and do not have any Editor's Draft on purpose, but still have a repository that we want to keep track of. One example is `device-memory-1`:
 https://www.w3.org/TR/device-memory-1/

This update adds a mechanism that allows setting `repository` to `null` in the `specs.json` file, to prevent the computation of the repository for the entry.

@dontcallmedom I'm actually wondering whether it wouldn't be preferable to just relax the test "there can be no two entries with the same repository + sourcePath" as the data isn't really incorrect.

### DIFF
--- a/schema/definitions.json
+++ b/schema/definitions.json
@@ -63,43 +63,36 @@
       "enum": ["w3c", "spec", "ietf", "whatwg", "iso"]
     },
 
-    "nightly": {
-      "type": "object",
-      "properties": {
-        "url": { "$ref": "#/$defs/url" },
-        "status": {
-          "type": "string",
-          "enum": [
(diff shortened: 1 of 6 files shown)


dontcallmedom:
I think it's "a bit" incorrect, in the sense that detecting activity on the repo wouldn't really be a signal of activity on the said spec, so I think the `null` approach feels like better semantics

tidoust:
Good point about activity signals. Go for `null`, then. I updated the JSON schemas accordingly.

dontcallmedom (approved):

PR #2512, opened 2026-05-29
New `--skip-specs` option to `build-index` script

### LINKED ISSUE #2509: Option to temporarily skip a spec during build
The repository for the [WASM Content Security Policy proposal](https://github.com/WebAssembly/content-security-policy) got deleted by accident a couple of days ago. The repository is being re-created but GitHub Pages are not yet setup again as I write this issue.

Build fails when a spec URL returns a 404. That's a **good thing** because it alerts us that something is wrong. It would still be great to have a way to skip building a spec once we're confident that the problem is a temporary one, so that other updates may propagate. Build would just reuse the last published info in such cases.

tidoust:
Fixes #2509 by introducing a new option to the build script that takes a comma-separate list of URLs as value and makes the build script reuse the last published info for these entries without trying to gather information for the spec.

The option should only be used to force a build when we run in a situation where one of the specs has a temporary hiccup (as illustrated by the WASM proposal repository deleted by accident), to be able to propagate other changes.

### DIFF
--- a/src/determine-filename.js
+++ b/src/determine-filename.js
@@ -17,9 +17,9 @@ export default async function (url) {
   }
 
   // RFC-editor HTML rendering
-  const rfcMatch = url.match(/\/rfc\/(rfc[0-9]+)$/);
+  const rfcMatch = url.match(/\/info\/(rfc[0-9]+)$/);
   if (rfcMatch) {
-    return rfcMatch[1] + '.html';
+    return null;
   }
 
   // Make sure that url ends with a "/"
--- a/.github/workflows/build-skip-iso.yml
+++ b/.github/workflows/build-skip-iso.yml
@@ -6,6 +6,11 @@ on:
     # (see build.yml workflow for Saturday and Sunday)
     - cron: '10 */6 * * MON-FRI'
   workflow_dispatch:
+    inputs:
+      skipSpecs:
+        description: "Spec URLs to skip (CSV string)"
+        required: false
+        type: string
 
 jobs:
   fetch:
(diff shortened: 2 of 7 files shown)


src/determine-filename.js
```diff
@@ -17,9 +17,9 @@ export default async function (url) {
   }
 
   // RFC-editor HTML rendering
-  const rfcMatch = url.match(/\/rfc\/(rfc[0-9]+)$/);
+  const rfcMatch = url.match(/\/info\/(rfc[0-9]+)$/);
   if (rfcMatch) {
-    return rfcMatch[1] + '.html';
+    return null;
```
tidoust, line 22: OK, these changes should have been part of #2511 as well. Not a big deal in any case, the regular expression no longer matched any RFC, so the function effectively already returns null for RFCs. This update just makes it return null sooner.

dontcallmedom (approved):
LGTM, but shouldn't the possibility of adding the skip specs parameter be integrated in the workflow file as well? or do you expect this would be only used in a local checkout?

tidoust:
> LGTM, but shouldn't the possibility of adding the skip specs parameter be integrated in the workflow file as well? or do you expect this would be only used in a local checkout?

Yes, that's a good idea. I was more targeting local checkout only. I made the change. Hopefully I didn't introduce new bugs in the workflows ;)

tidoust:
> Hopefully I didn't introduce new bugs in the workflows ;)

And I did, of course :-D
Fixed in https://github.com/w3c/browser-specs/commit/4f85c8efcf75572825e29baa658dc3a453d9baae

Issue #2536, opened 2026-06-15
draft-ietf-httpapi-idempotency-key-header-07 is expired and archived


Elchi3:
It seems that https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-idempotency-key-header/#name-the-idempotency-key-http-re is expired and archived. 

It was originally requested in https://github.com/w3c/browser-specs/issues/2199, however, in the meantime (since April 2026) the spec expired and Firefox is also disabling the implementation again https://bugzilla.mozilla.org/show_bug.cgi?id=2040423

Should we mark it with "standing: discontinued"?

tidoust:
> Should we mark it with "standing: discontinued"?

Yes, that would be the right thing to do.

It would also be good to dig into the code that sets the `nightly.status` property for IETF drafts. I suspect it just sets it to `"Editor's Draft"` by default, but we should be able to trap the "expired and archived" status and turn it into a `"Discontinued Draft"` value, which would then make the `standing` switch to `"discontinued"` automatically.

tidoust:
Testing locally, detecting the "Expired" state automatically would turn 3 entries to discontinued:
- [The Idempotency-Key HTTP Header Field](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-idempotency-key-header)
- [Client Hint Reliability](https://datatracker.ietf.org/doc/html/draft-davidben-http-client-hint-reliability)
- [Cookies Having Independent Partitioned State specification](https://datatracker.ietf.org/doc/html/draft-cutler-httpbis-partitioned-cookies)

That may be good, I do not know to what extent the "Expired" state signals "discontinued" work in practice.

Elchi3:
> * [Client Hint Reliability](https://datatracker.ietf.org/doc/html/draft-davidben-http-client-hint-reliability)

In BCD, we still reference https://www.ietf.org/archive/id/draft-davidben-http-client-hint-reliability-03.html#name-the-critical-ch-response-he which is also under an "archive" URL. Looking at https://github.com/davidben/http-client-hint-reliability there doesn't seem to be much activity on the standardization process...
 
> * [Cookies Having Independent Partitioned State specification](https://datatracker.ietf.org/doc/html/draft-cutler-httpbis-partitioned-cookies)

[...]

PR #2548, opened 2026-06-25
Auto-flag expired IETF drafts as discontinued drafts

tidoust:
See discussion in #2536. The code that fetches information about an IETF draft now leverages the `state` property in the JSON to set the status to `"Discontinued Draft"` for `"Expired"` IETF drafts.

This will also automatically switch the standing of the underlying spec in browser-specs to `"discontinued"`.

The update will impact the 3 identified expired IETF drafts mentioned in the issue:
- The Idempotency-Key HTTP Header Field
- Client Hint Reliability
- Cookies Having Independent Partitioned State specification


Elchi3 (approved):
Thanks @tidoust 👍

PR #2593, opened 2026-08-07
Soft Navigations is now "good"

tunetheweb:
The Soft Navigations spec was included [unflagged in Chrome 151](https://chromestatus.com/feature/5144837209194496) so I believe it should no longer be "pending".

It's still in WICG so not yet a "standard", but is up to date and being maintained.


tidoust:
Yes, seems good, thanks for the update!

PR #2617, opened 2026-08-25
Setup Node.js after checkout in workflows

tidoust:
Workflows now read the version of Node.js to use from the `.node-version` file. A couple of them attempted to setup Node.js before checking out the repository, which cannot work. This update makes sure the checkout action runs before the setup-node action in all workflows.


tidoust:
Puppeteer install error reported in tests is unrelated (and hopefully transient).

PR #2640, opened 2026-09-18
Update URLs that have moved

tabatkins:
In addition to the changes in the PR, all of the following have their TR url listed in the data, and an old ED url is extracted from the document. All of them now redirect to a `/aria/` URL and should be updated, but I'm not sure exactly how the data handles that.
* `https://www.w3.org/TR/accname-1.2/`
* `https://www.w3.org/TR/core-aam-1.2/`
* `https://www.w3.org/TR/dpub-aam-1.1/`
* `https://www.w3.org/TR/dpub-aria-1.1/`
* `https://www.w3.org/TR/graphics-aam-1.0/`
* `https://www.w3.org/TR/graphics-aria-1.0/`
* `https://www.w3.org/TR/html-aam-1.0/`
* `https://www.w3.org/TR/svg-aam-1.0/`


specs.json
```diff
@@ -645,7 +645,7 @@
   "https://w3c-cg.github.io/web-nfc/",
   "https://w3c-fedid.github.io/login-status/",
   "https://w3c.github.io/at-driver/",
-  "https://w3c.github.io/css-aam/",
+  "https://w3c.github.io/aria/css-aam/",
```
tidoust, line 648: As things stand, the code that builds the final list does not understand this pattern and will happily claim that the shortname of the spec is `aria`. The build code needs to be updated to compute the right shortname.

specs.json
```diff
@@ -806,9 +806,9 @@
     "url": "https://webaudio.github.io/web-speech-api/",
     "shortname": "speech-api"
   },
-  "https://webbluetoothcg.github.io/web-bluetooth/",
+  "https://bluetooth.spec.whatwg.org/",
```
tidoust, line 809: This and the other changes are going to change the shortname of the spec from `web-bluetooth` to `bluetooth`. We need to record the former name in a `formerNames` property.

specs.json
```diff
@@ -1024,10 +1024,10 @@
   },
   "https://wicg.github.io/webcrypto-modern-algos/",
   "https://wicg.github.io/webcrypto-secure-curves/",
-  "https://wicg.github.io/webhid/",
+  "https://hid.spec.whatwg.org/",
```
tidoust, line 1027: Same comment as above regarding change of shortname.

specs.json
```diff
@@ -1024,10 +1024,10 @@
...
   "https://wicg.github.io/webpackage/loading.html",
   "https://wicg.github.io/webrtc-diagnostic-logging/",
-  "https://wicg.github.io/webusb/",
+  "https://usb.spec.whatwg.org",
```
tidoust, line 1030: Same comment as above regarding change of shortname.

tidoust (changes requested):
Thanks, @tabatkins! The updates are good, but:

1. The new URLs for ARIA are going to confuse the code that produces the final list. I'll look into updating the code to handle that pattern.
2. Some of the WHATWG URLs produce new shortnames for the spec. The former shortname should be recorded in the `formerNames` property so that consumers who need to track such updates can find the new entry in the list from the former shortname. I'm also happy to handle that.

The tests fail because the entries in the list are sorted by URL so the updates need to move elsewhere.

tidoust:
> In addition to the changes in the PR, all of the following have their TR url listed in the data, and an old ED url is extracted from the document.

browser-specs takes that data from the W3C API, which gets that data from the specs themselves. The /TR specs should be re-published to target the new Editor's Drafts. In the meantime, we should be able to force the URL of the Editor's Draft in the API. Cc @deniak.

tidoust:
I split these updates into more atomic updates and updated the underlying build logic accordingly. The list should now have the right information for the ARIA specs that have not yet been published to /TR, and for the Peripheral APIs.

What's still pending is updating the URLs of the Editor's Drafts of the ARIA documents already published on the Recommendation track.

deniak:
> > In addition to the changes in the PR, all of the following have their TR url listed in the data, and an old ED url is extracted from the document.
> 
> browser-specs takes that data from the W3C API, which gets that data from the specs themselves. The /TR specs should be re-published to target the new Editor's Drafts. In the meantime, we should be able to force the URL of the Editor's Draft in the API. Cc @deniak.

I just updated the EDs in the W3C API!

tidoust:
Updates shipped in new version web-specs ([v4.15.0](https://www.npmjs.com/package/web-specs/v/4.15.0)). Closing as addressed.

